PR #4, opened 2017-02-28
fastMap functions

Adriel-M:
Added three functions related to fastMap; fastMap, fastMapUpdate, fastMapGenerate. Included unit tests for all three functions and integrations tests between them.


Adriel-M:
I have done all the requested changes. Wasn't aware of the NA; was trying to insert NULL into a character vector. Regarding the path, I just simplified it by default saving in working director and just checking of the supplied outputName contains a .rds extension, program will continue but a warning will be raised.

Regarding fastMapSanity, keys and value can contain hyphens which the regex detects.

PR #7, opened 2017-03-02
Committing: Test_Data.R, FINDSUBv1_2.R

whitleyo:
Here are the following
1. Findsubv1_2.R: FINDSUB function defined in this script
2. Test_Data.R: sources the FINDSUBv1_2.R file, and also defines a hypothetical network of 9 proteins
Edge and vertex data frames are formatted with expected attributes, and are converted into an iGraph object

3. FINDSUB_Test1.R:FINDSUB_Test3.R files test the function using different thresholds for minOrd and Thresh (minimum number of vertices per subnetwork and edge weight threshold, respectively), and in FINDSUB_Test4 changing edge weights to see if subnetworks formed would change relative to results of test 1 as expected. FINDSUB_Test5 tests that entering data of incorrect type for the function inputs results in stoppage of the function.

Note: You will have to change filepaths for the TestData and FINDSUB_TestX files (X refers to any number test). Basically, my approach was to source the function in the TestData file, and then source the TestData file in every FINDSUB_TestX.R file.


whitleyo:
Pushed the changes requested.

________________________________

hyginn:
Superseded

PR #9, opened 2017-03-13
added importnet.multinet

iyifan:
added importnet.multinet


iyifan:
added importnet.multinet

PR #19, opened 2017-03-17
Add importSNV.TCGA file

Pruthv1:
* added importSNV.TCGA file with comments about workflow and how the data will be handled


R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
+# importNet.TCGA.R
+
+#' Import MAF files from source and coverts them to rSNV files.
```
hyginn: *convert

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+#'
+#' \code{importSNV.TCGA} imports maf, creates rSNV, returns rSNV
+#'
+#' @section MAF file:...
+#'
+#' @param fMAF vector of MAF file names
+#' @param rSNV new rSNV file name to be created or the one to be merged with
+#' @param silent Controls whether output to console should be suppressed. FALSE
+#'   by default.
+#' @param noLog Controls whether writing the result to the global logfile is
```
hyginn: ```writeLog = TRUE```

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+#'   ## @examples ## \dontrun{ ## importNet.STRING(IN, OUT) ## }
+#' @export
+importSNV.TCGA <- function(fMAF,
+                             rSNV,
+                             silent = FALSE,
+                             noLog = TRUE) {
+
+    # ==== PARAMETERS ==========================================================
+
+    # NL <- .PlatformLineBreak()
```
hyginn: Needed

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    # cutoffTypes <- c("xS", "xQ", "xN")
```
hyginn: 32-38: not needed

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    #
+    # data   <- unlist(strsplit(tmp[2], " "))
+
+
+    # ==== VALIDATIONS =========================================================
+
+    # General parameter checks
+    cR <- character()
+    cR <- c(cR, .checkArgs(fMAF,        like = c("FILE_E"),   checkSize = TRUE))
+    cR <- c(cR, .checkArgs(rSNV,      like = "FILE_E",        checkSize = TRUE))
```
hyginn: The file does not exist at this point.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    cR <- c(cR, .checkArgs(silent,       like = logical(1), checkSize = TRUE))
+    cR <- c(cR, .checkArgs(writeLog,     like = logical(1), checkSize = TRUE))
+
+    if(length(cR) > 0) {
+        stop(cR)
+    }
+
+    #  Validate the vector of MAF file names parameter
+    #  Validate rSNV as an existing file or create new file
+
```
hyginn: If it exists, load it here.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    # ==== READ DATA ===========================================================
+
+    # reading data will occur in multiple steps to avoid storing too much data on
+    # the disk drive and cause overload when handeling large files
+
+    # step 0: run terminal comand to gzip all folders and subfolders and remove
```
hyginn: no. why?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    #         unwanted files like stdout, stdin, manifest etc..
+    # step 1: go through diseases for which data is available and open file
```
hyginn: No. Accept the vector of input files as is. Don't second guess the user here.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    # step 2: read lines from file and process data as indicated in "EXTRACT" below
```
hyginn:
```
Read into a temporary dataframe in one step using readr::read_delim
Read only desired columns. 
Typecheck columns after read (typecheck elements in first row). Error handling in case typecheck fails (eg. would a string in a numeric column cause the column to be converted to char? Or is readr::read_delim safe if the column type is specified?)
Process data.
Append temp data frame to old dataframe.
delete temp dataframe
explicit file close/open are unnecessary.

File: R/importSNV.TCGA.R [hunk 10]
```diff
@@ -0,0 +1,156 @@
...
+        stop(cR)
+    }
+
+    #  Validate the vector of MAF file names parameter
+    #  Validate rSNV as an existing file or create new file
+
+    # ==== READ DATA ===========================================================
+
+    # reading data will occur in multiple steps to avoid storing too much data on
+    # the disk drive and cause overload when handeling large files
```
hyginn: Does readr read directly from compressed files? If so, you can make use of that (the user will pass in a ".gz" filename. If not, must ensure the file is a textfile. If it is not, skip the import.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    cR <- c(cR, .checkArgs(fMAF,        like = c("FILE_E"),   checkSize = TRUE))
+    cR <- c(cR, .checkArgs(rSNV,      like = "FILE_E",        checkSize = TRUE))
+    cR <- c(cR, .checkArgs(silent,       like = logical(1), checkSize = TRUE))
+    cR <- c(cR, .checkArgs(writeLog,     like = logical(1), checkSize = TRUE))
+
+    if(length(cR) > 0) {
+        stop(cR)
+    }
+
+    #  Validate the vector of MAF file names parameter
```
hyginn: Validate for readable file-type, then check contents by reading header and first line.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    # iCol <- which(net == header)  # column that contains the requested net
+    #
+    # data   <- unlist(strsplit(tmp[2], " "))
+
+
+    # ==== VALIDATIONS =========================================================
+
+    # General parameter checks
+    cR <- character()
+    cR <- c(cR, .checkArgs(fMAF,        like = c("FILE_E"),   checkSize = TRUE))
```
hyginn: This will fail if the input is a vector. You can do:
```
cR <- c(cR, .checkArgs(fMAF,
                    like = rep("FILE_E", length(fMAF)),
                    checkSize = TRUE))
```

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    cR <- c(cR, .checkArgs(rSNV,      like = "FILE_E",        checkSize = TRUE))
+    cR <- c(cR, .checkArgs(silent,       like = logical(1), checkSize = TRUE))
+    cR <- c(cR, .checkArgs(writeLog,     like = logical(1), checkSize = TRUE))
```
hyginn: This is correct. Your @param and function signature is not.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    # 10 Variant_type
+    # 11 Reference_allele
+    # 12 Tumor_Seq_Allele1
+    # 13 Tumor_Seq_Allele2
+    # 16 Tumor_Sample_Barcode
+    # 33 Tumor_Sample_UUID
+
+    # read each sample from file and parse the columns listed above and write them
+    # to the rSNV file.
+    # incase of missing value add index of sample entry in rSNV to vector 'missingValues'
```
hyginn: What's a "missing value"?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    # when we reach the last sample close file and open new file.
+
+    # ==== "HANDLE" - MISSING VALUES =============================================
+
+    # Incase we encounter missing attributes for samples
```
hyginn: No. Unnecessary. Just record the number of dropped rows (if any) write a message to console and add a note to the log message.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+        myCall[5] <- sprintf("writeLog = %s)", as.character(writeLog))
+        myCall <- paste0(myCall, collapse = "")
+
+        # indicate input object name(s)
+        #   (NA for this importSNV.TCGA())
+
+        # Record progress information
+        myNotes <- character()
+        myNotes <- c(myNotes, sprintf("Files processed - ", nDone))
+        nyNotes <- c(myNotes, sprintf("Files pending - ", nPend))
```
hyginn: Under what circumstances could there be pending files?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+
+    if(writeLog) {
+
+        myTitle <- "importSNV.TCGA"
+
+        # Compile function call record
+        myCall <- character()
+        myCall[1] <- "importSNV.TCGA("
+        myCall[2] <- sprintf("rMAF = \"%s\", ", fName)
+        myCall[3] <- sprintf("rSNV = \"%s\", ", net)
```
hyginn: There is no "net" in your arguments!

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    # ==== WRITE LOG ===========================================================
+
+    if(writeLog) {
+
+        myTitle <- "importSNV.TCGA"
+
+        # Compile function call record
+        myCall <- character()
+        myCall[1] <- "importSNV.TCGA("
+        myCall[2] <- sprintf("rMAF = \"%s\", ", fName)
```
hyginn: Not correct: fName could be a vector. Need to handle multiples.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+        myCall <- paste0(myCall, collapse = "")
+
+        # indicate input object name(s)
+        #   (NA for this importSNV.TCGA())
+
+        # Record progress information
+        myNotes <- character()
+        myNotes <- c(myNotes, sprintf("Files processed - ", nDone))
+        nyNotes <- c(myNotes, sprintf("Files pending - ", nPend))
+        myNotes <- c(myNotes, sprintf("Progress in percentage =
```
hyginn: Progress?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+                                      Size of Data Processed/Total Size of Data",
+                                      nComplete))
+        for (i in missingEntry) {
```
hyginn: No. Just drop ... also: these could be hundreds of thousands if there is a problem with the file!!!

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+                                          K of rSNV", K))
+        }
+
+        # indicate output object name(s)
+        myOutput = c("rSNV")
+
+        # send info to log file
+        logEvent(eventTitle = myTitle,
+                 eventCall = myCall,
+#                input = character(),
```
hyginn: If your old rSNV already exists, **that** will be input.But not if a new object is created, then input information is only the filenames.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+        # Compile function call record
+        myCall <- character()
+        myCall[1] <- "importSNV.TCGA("
+        myCall[2] <- sprintf("rMAF = \"%s\", ", fName)
+        myCall[3] <- sprintf("rSNV = \"%s\", ", net)
+        myCall[4] <- sprintf("silent = %s, ", as.character(silent))
+        myCall[5] <- sprintf("writeLog = %s)", as.character(writeLog))
+        myCall <- paste0(myCall, collapse = "")
+
+        # indicate input object name(s)
```
hyginn: See below ...

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+    # ==== "HANDLE" - MISSING VALUES =============================================
+
+    # Incase we encounter missing attributes for samples
+    # Add the sample entry to the rSNV file but record the line number
+    # or the index of that entry.
+    # Make a vector of all entries with missing values
+    # Report all missing entries in the log Below.
+
+    # can also have and option to drop all entries with missing values
+
```
hyginn: Don't forget to add metadata to the output rSNV.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,156 @@
...
+        # send info to log file
+        logEvent(eventTitle = myTitle,
+                 eventCall = myCall,
+#                input = character(),
+                 notes = myNotes,
+                 output = myOutput)
+    }
+
+    return(gG)
```
hyginn: Do you **always** return the table, or only if you did not write it to file?

Also, this is ***not*** gG.

hyginn:
I think this is oK - after you consider the notes (especially making the logic of handling multiple input files, and of creating/appending the rSNV consistent.) So go ahead and design the tests. (In a new file).

hyginn:
mustnotpushwrongbutton

R/importSNV.TCGA.R
Pruthv1: I meant pending as in, what MAF files have been processed and what remain from input data
But now that I think about it, it seems irrelevant

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,148 @@
...
+#'   ## @seealso \code{\link{fastMap}} fastMap() is used internally to map ENSP
+#'   IDs to gene symbols.
+#'
+#'   ## @examples ## \dontrun{ ## importNet.STRING(IN, OUT) ## }
+#' @export
+importSNV.TCGA <- function(  fMAF,
+                             rSNV,
+                             silent = FALSE,
+                             writeLog = TRUE,
+                             na.rm = TRUE) {
```
hyginn: Keep silent/writelog as last arguments for consistency.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,148 @@
+# importNet.TCGA.R
```
hyginn: Change filename to ```importSNV.TCGA.R``` - it's not a network import.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,148 @@
...
+    # 10 Variant_type
+    # 11 Reference_allele
+    # 12 Tumor_Seq_Allele1
+    # 13 Tumor_Seq_Allele2
+    # 16 Tumor_Sample_Barcode
+    # 33 Tumor_Sample_UUID
+
+    # read each sample from file and parse the columns listed above and write them
+    # to the rSNV file.
+    # write missing/dropped to log HERE
```
hyginn: No.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,148 @@
...
+    # incase of missing value add index of sample entry in rSNV to vector 'missingValues'
```
hyginn: No.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,148 @@
...
+    # when we reach the last sample close file and open new file.
+
```
hyginn: ```open()```/```close()``` not necessary. 
Also: remember you are no longer processing in streaming mode.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,148 @@
...
+    # ==== "HANDLE" - MISSING VALUES =============================================
+
+    # Incase we encounter missing attributes for samples
+    # Add the sample entry to the rSNV file but record the line number
+    # or the index of that entry.
+    # Make a vector of all entries with missing values
+    # Report all missing entries in the log Below.
+
+    # can also have and option to drop all entries with missing values
```
hyginn: Yes.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,148 @@
...
+    # 16 Tumor_Sample_Barcode
+    # 33 Tumor_Sample_UUID
+
+    # read each sample from file and parse the columns listed above and write them
+    # to the rSNV file.
+    # write missing/dropped to log HERE
+    # incase of missing value add index of sample entry in rSNV to vector 'missingValues'
+    # when we reach the last sample close file and open new file.
+
+    # ==== "HANDLE" - MISSING VALUES =============================================
```
hyginn: You have to be more explicit about the expected function behaviour for this section: what **exactly** is a "missing value". NA? Something else too? Empty cells? Does ```readr::read_delim()``` turn empty cells into NA? Etc. If you are not 100% specific, you can't define what you will test for and how your tests should check whether the function works correctly. Not to go overboard here, I don't think we need to check for every possible thing that could go wrong - we'll get to erroneous input when we go over the test design - but the way this is written I am not sure what you mean, so as the code reviewer I need more guidance.

hyginn:
I've added some more short comments - have a look, change if needed, and then let's get on with designing the tests.

hyginn:
I see a number of commits: please write an explicit comment here that it is ready for review whenever it is.
Thanks,
Boris

Pruthv1:
Test file is ready for review.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,73 @@
...
+    # if rSNV parameter is provided, rSNV file does not exist
+    # if rSNV parameter is provided, but is NULL
+    # if rSNV parameter is provided, but is non-string value
+    # if rSNV is not provided, new file could not be created (file name conflict)
+    # if na.rm is not given logical input
+    # if silent is not given logical input
+    # write.log is not given logical input
+})
+
+test_that("check various forms of input parameters work correctly", {
```
hyginn: I don't see that you can check that without running the actual code. I think these tests need to be merged with the next block ("works correctly")

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,73 @@
...
+    # check if manual output = module output
+})
+
+
+test_that("a corrupt input does not lead to corrupted output", {
+    # if fMAF file open does not follow the 2.4 or agreed upon version either STOP or DROP file
+    # if fMAF file open does not have the required columns at the
+    #           right column number STOP or DROP file
+    # if rSNV is provided, check if number of columns in rSNV is worng
+    # if rSNV is provided, check when header labels are wrong
```
hyginn: Not just wrong number - but column names too need to be identical.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,73 @@
...
+})
+
+test_that("na.rm works as intended", {
+    # if fMAF file has N/A values in rows. if na.rm == TRUE. check if the row
```
hyginn: Careful:  "N/A" not applicable - not an R reserved word. ```NA``` - not available. Reserved word in R to signify missing data. Also: "NA" != ```NA```.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,73 @@
...
+    #           is dropped and not included in rSNV
+    # if fMAF file has N/A values in rows. if na.rm == FALSE check if the row
```
hyginn: This is the same as 46. Probably unintended ...

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,73 @@
...
+    #           is not dropped and included in rSNV
+
+})
+
+test_that("silent and write works as intended", {
+    # if silent=TRUE, check if output to console is supressed
```
hyginn: Actually we didn't check that in the other codes - but you are right, we should. Use ```capture.output()``` like so:

```
testF <- tempfile()
capture.output(Do.A.Thing(silent = TRUE), file = testF)
expect_equal(length(readLines(testF)), 0)
expect_true(file.remove(testF))
```

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,73 @@
...
+    # if silent=FALSE, check if output to console is done correctly
+    # if writeLog=TRUE, check if log file is created
+    # if writeLog=TRUE, check if logs are written to it correctly
+    # if writeLog=FALSE, check that no log file is created
+    # if writeLog=FALSE, cmake sure no log files are written
+})
+
+test_that("test that columns contains valid values", {
+    # (not sure if this is necessary, do we just assume input data is correct?)
```
hyginn: NO

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,73 @@
...
+})
+
+test_that("silent and write works as intended", {
+    # if silent=TRUE, check if output to console is supressed
+    # if silent=FALSE, check if output to console is done correctly
+    # if writeLog=TRUE, check if log file is created
+    # if writeLog=TRUE, check if logs are written to it correctly
+    # if writeLog=FALSE, check that no log file is created
+    # if writeLog=FALSE, cmake sure no log files are written
+})
```
hyginn: Check existence and format of metadata in resulting RDS

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,73 @@
...
+
+test_that("test that columns contains valid values", {
+    # (not sure if this is necessary, do we just assume input data is correct?)
+    # check if we are using the correct MAF versions like 2.4 or lower etc.
+    # check if variant_class contains valid values from known list
+    # check if vairant_type contains valid values from known list
+    # check UUID format
+    # check chromosome value
+    # check start position and end are integers
+    # check tumor sample barcode is correct format
```
hyginn: No, not that - we are not using it. Only check things that influence or go into the output (Also not 67).

hyginn:
Good. Some comments added, but please proceed with implementing the tests. Include the setup/teardown blocks from the most recent version of ```tests/testthat/testDummy.R```.

Pruthv1:
tests Ready for review.

Pruthv1:
had problems with my fork version, hence the multiple commits. Ready for review!

hyginn:
Pruthvi - I won't be able to check this in at all, there are dozens of changes in files that you did not edit and I can't be sure that that won't undo later changes in working code. You need to remove your files locally from your repo, delete your fork and RStudio project, create a new fork and project, then add your new files back.

I had asked whether people need help with the fork-update process, no one responded, so please do this the correct way.

Once again: this pull request can't be merged.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+    expect_error(importSNV(writeLog = 1))
+    expert_error(importSNV(writeLog = NULL))
+    expert_error(importSNV(wrtieLog = logical()))
+    expert_error(importSNV(writeLog = c(TRUE, TRUE)))
+})
+
+test_that("a sane input gives an expected output with differnt input parameters", {
+
+    # make small input (testIn)
+    testIn <- "~/Desktop/rete/testIn"
```
hyginn: That's not portable.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+    # make small output for that input manually (testOut)
+    write("# metadata \n# 1 \n# 2  \n# 3 \n# 4 \n# 5 \n# metadata END\n", "testOut")
```
hyginn: Why don't you just create a data frame trueOut? Also you'll need to turn stringsAsFactors off.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+
+    # run the module on only fMAF as parameter (moduleOut)
+    moduleOut <- importSNV(c("testIn"))
+    # check if manual output = module output
+    # skip 7 for metadata lines just for now
+    module <- read.csv("moduleOut", skip=7, sep = "\t")
+    all.equal(trueOut, module)
+
+    # test input with rSNV as input
+    testInPre <- "~/Desktop/rete/testInPre"
```
hyginn: Not portable.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
+# testImportSNV.TCGA.R
```
hyginn: Generally: turn writeLog off in all tests except in one block where you specifically test log output. Make sure there is no log file before you test, and after your tests delete the one you created. Log files are being appended to, and if you don't take care, your tests will fail because of old contents.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+    # if rSNV is provided, check if number of columns in rSNV is worng
+    expect_error(importSNV(c("testIn")), "testOutINVALIDcolumns")
+    # if rSNV is provided, check when header labels are wrong
+    expect_error(importSNV(c("testIn")), "testOutINVALIDheaders")
+})
+
+test_that("silent and write works as intended", {
+    # if silent=TRUE, check if output to console is supressed
+    testF <- tempfile()
+    capture.output(Do.A.Thing(silent = TRUE), file = testF)
```
hyginn: Might want to change the function name here :-)

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+    expect_true(file.remove(testF))
+
+    # if writeLog=TRUE, check if log file is created
+
+    # if writeLog=TRUE, check if logs are written to it correctly
+
+})
+
+
+test_that("test that columns contains valid values", {
```
hyginn: Better to move this to the valid input -> valid output section.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+    # check if variant_class contains valid values from known list
+    ### will finalize these tests once import module is implemented ###
+
+    classColumn <- addColumn("testOut")
```
hyginn: I'm not completely sure what you are doing here - but this looks like a test that actually belongs in your function. Rather than iterate through the choices, provide one test input that contains each of the options, and test once that your functions passes all of them. Then provide *one* wrong keyword, and confirm that the import fails.

Also, use ```%in%``` to check whether some observed keywords are in a list, not iteration. Something like:
```
myVclasses <- unique(data$variant.class)
if (! all(myVclasses %in% knownVclasses)) { ... that's an error.
```

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+    if (i == lenght(typeColumn)) {
+        print("All values Valid")
+    } else {
+        print("Invalid values in type column!")
+    }
+
+    # check chromosome value
+    chrColumn <- addColumn("testOut")
+    typeChr <- c("chr1","chr2","chr3","chr4","chr5","chr6","chr7","chr8","chr9","chr10","chr11","chr12","chr13",
+                 "chr14","chr15","chr16","chr17","chr18","chr19","chr20","chr21","chr22","X","Y","Z")
```
hyginn: Z?
And what about "mt"?

hyginn:
I have added some comments to your tests file.

tests/testthat/testImportSNV.TCGA.R
Pruthv1: they are lowercase. oh and my mistake "Z" is "M"

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+    moduleOut <- importSNV(c("testIn"))
+    # check if manual output = module output
+    # skip 7 for metadata lines just for now
+    module <- read.csv("moduleOut", skip=7, sep = "\t")
+    all.equal(trueOut, module)
+
+    # test input with rSNV as input
+    testInPre <- "~/Desktop/rete/testInPre"
+    # test output for rSNV as input
+    sample1 <- c("VWA5B1", "chr1",	20319393,	20319393,	"+",	"Missense_Mutation",	"SNP", 	"C",	"C",	"T", "TCGA-QQ-A8VF-01A-11D-A37C-09", "2109690d-e8b1-4b5c-9a39-21af273324cf")
```
RicardoRam: 80 character per line limit

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+    # check if vairant_type contains valid values from known list
+    typeColumn <- addColumn("testOut")
+    typeV <- c("SNP", "DNP", "TNP", "ONP", "INS", "DEL", "Consolidated2")
+    out.V <- TRUE
+    i <- 1
+    while(out.V == TRUE & i <= length(classColumn)){
+        out.V <- is.element(typeColumn[i], typeV)
+        i <- i + 1
+    }
+    if (i == lenght(typeColumn)) {
```
RicardoRam: length*

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+    classV <- c("Frame_Shift_Del", "Frame_Shift_Ins", "In_Frame_Del", "In_Frame_Ins", "Missense_Mutation",
+                "Nonsense_Mutation", "Silent", "Splice_Site", "Translation_Start_Site", "Nonstop_Mutation",
+                "3'UTR", "3'Flank", "5'UTR", "5'Flank", "IGR1" , "Intron", "RNA", "Targeted_Region")
+    out.C <- TRUE
+    i <- 1
+    while(out.C == TRUE & i <= length(classColumn)){
+        out.C <- is.element(classColumn[i], classV)
+        i <- i + 1
+    }
+    if (i == lenght(classColumn)) {
```
RicardoRam: length*

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,208 @@
...
+    all.equal(trueOutPre, modulePre)
+
+    # run the module on fMAF, rSNV, silent and writeLog as parameters
+    moduleOutPre <- importSNV(c("testInPre"), "module02", silent=TRUE, writeLog=TRUE)
+    modulePre <- read.csv("moduleOutPre", skip=7, sep = "\t")
+    # add to prexisting rSNV file
+    all.equal(trueOutPre, modulePre)
+
+    # test input with rSNV as input and NA values
+    testInNA <- "~/Desktop/rete/testInNA"
```
RicardoRam: also not portable

PR #20, opened 2017-03-20
remake fork, readded importSNV.TCGA and tests

Pruthv1:
* remake of fork
* changes added to test file
* changes added and implemented parts of importSNV.TCGA


Pruthv1:
* Ready for review!

hyginn:
Can we close #19? You can still access comments via the linkt o "Closed" requests ...

Pruthv1:
yes #19 can be closed. Everything is in #20 now.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
+# importSNV.TCGA.R
+
+#' Import MAF files from source and converts them to rSNV files.
+#'
+#' \code{importSNV.TCGA} imports list of maf file names, creates rSNV or adds to
+#' existing rSNV file, return rSNV file
```
hyginn: *returns

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
+# importSNV.TCGA.R
+
+#' Import MAF files from source and converts them to rSNV files.
+#'
+#' \code{importSNV.TCGA} imports list of maf file names, creates rSNV or adds to
```
hyginn: *imports data from one or more files in maf format, ...

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+#' existing rSNV file, return rSNV file
+#'
+#' @section MAF file:...
```
hyginn: ToDo

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+#'
+#' @param fMAF vector of MAF file names
+#' @param rSNV rSNV file to merged new MAF data with
+#' @param na.rm remove all rows with NA values as attributes and drop from DF
```
hyginn: what does "as attributes" mean?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+#' @param silent Controls whether output to console should be suppressed. FALSE
+#'   by default.
+#' @param writeLog Controls whether writing the result to the global logfile is
+#'   enabled. TRUE by default.
+#' @return rSNV file name containg data from all MAF files
```
hyginn: ``` N/A. Function is invoked for its side-effect of saving an rSNV object in RDS format to the requested filename.```

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+#' @family
+#'
+#'   ## @seealso \code{\link{importX}} importX() is used to import ...
+#'
+#'   ## @examples ## \dontrun{ ## importSNV.TCGA(IN, OUT) ## }
+#' @export
+
+importSNV.TCGA <- function(  fMAF,
+                             rSNV = NULL,
+                             na.rm = FALSE,
```
hyginn: Don't set it to NULL by default. Just give it no default value and test for ```missing()```.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+    myNotes <- character()
+
+    # ==== PARAMETERS ==========================================================
+
+    # NL <- .PlatformLineBreak()
+    # # Read header and one contents line
+    # tmp <- readLines(fName, n = 2)
+
+    # ==== VALIDATIONS =========================================================
+
```
hyginn:
```
if (missing(fName)) {
    stop("The filename of the input file must be specified as \"fName\".")
}
if (missing(rSNV)) { ...
```

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+                                                             checkSize = TRUE))
+    cR <- c(cR, .checkArgs(silent,       like = logical(1), checkSize = TRUE))
+    cR <- c(cR, .checkArgs(writeLog,     like = logical(1), checkSize = TRUE))
+    cR <- c(cR, .checkArgs(na.rm,        like = logical(1), checkSize = TRUE))
+
+    if(length(cR) > 0) {
+        stop(cR)
+    }
+
+    meta <- list(type = "rSNV",
```
hyginn: ... add UUID

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+            return(TRUE)
+        } else {
+            return(FALSE)
+        }
+    }
+
+    mafHeaderCheck <- function(file) {
+        headValid <- TRUE
+        if (! read.table(file, header = FALSE, fill = NA, stringsAsFactors = FALSE,
+                         na.strings = " ", nrows=1)[,1] == "Hugo_Symbol") {
```
hyginn: Noooo. Don't define functions in here  :-)

Put the functions into the file separately, give them a function name that starts with a "." (e.g. .mafHeaderCheck) and remove the RoxygenHeader ( in particular the @export statement).

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+        if (! read.table(file, header = FALSE, fill = NA, stringsAsFactors = FALSE,
+                         na.strings = " ", nrows=1)[,33] == "Tumor_Sample_UUID") {
+            headValid <- FALSE
+        }
+        return(headValid)
+    }
+
+    # ==== Validata rSNV  =======================================
+
+    rSNVheadercheck <- function(file) {
```
hyginn: As above

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+            headValid <- FALSE
+        }
+        if (! read.table(file, header = FALSE, fill = NA, stringsAsFactors = FALSE,
+                         na.strings = " ", nrows=1)[,12] == "UUID") {
+            headValid <- FALSE
+        }
+        return(headValid)
+    }
+
+    if (missing(rSNV)) {
```
hyginn: Testing for missing parameters is usually the *first* thing in a function.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+        rSNV <- "MAFtoSNV"
+
+        # add metdata
+        for (name in names(meta)) {
```
hyginn: Again: no metadata here.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+            attr(rSNV, name) <- meta[[name]]
+        }
+
+        write("sym\tchr\tstart\tend\tstrand\tclass\ttype\taRef\ta1\ta2\tTCGA\tUUID",
+              rSNV, append=FALSE)
+
+        snvValidity <- TRUE
+    } else {
```
hyginn: You really don't need to read the entire old rSNV if it exists - you can just append to it after you confirm that the header is sane. use ```readLines()``` to read a single line, and check that.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+        rSNVf <- read.table(rSNV, header = TRUE)[,1]
+        rowsBefore <- length(rSNVf)
+        myNotes <- c("Number of rows originally in rSNV - ", rowsBefore)
+        snvValidity <- rSNVheadercheck(rSNV)
+    }
+
+    # ==== READ DATA ===========================================================
+    filesFinished <- 0
+    numFilesInvalid <- 0
+    for (i in fMAF) {
```
hyginn: This is not a good use of a progress bar, because the number of files is going to be small. In this case I would produce feedback like:
```
Processing "myNewTCGAdata-12345.maf" ... done. (1/3)
```
etc.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+
+        if (!silent) {
+            .pBar(i, length(fMAF))
```
hyginn: This is not a good use of a progress bar, because the number of files is going to be small. In this case I would produce feedback like:
```
Processing "myNewTCGAdata-12345.maf" ... done. (1/3)
```
etc.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+                         "chr14","chr15","chr16","chr17","chr18","chr19","chr20","chr21","chr22","chrX","chrY","chrM")
+            myVchr <- unique(chr_)
+            if (! all(myVchr %in% typeChr)) {
+                # drop file
+                chrValid <- FALSE
+            } else {
+                chrValid <- TRUE
+            }
+
+            if (classValid & typeValid & chrValid == FALSE) {
```
hyginn: :-)

```
if (! classValid && typeValid && chrValid) ...
```
(... or "&" if you must, but that's confusing in an if-statement, it just happens to work here.)

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+                # there was an invalid value
+                # report error in file and log
+                # drop file
+                # go to next
+                myNotes <- c(paste(i, " - File was dropped due to invalid attributes in column chr/class/type"))
```
hyginn: Ah. These are not "attributes" but data. Better to do the following:

instead of 
``` 
if (! all(myVclasses %in% classVknown)) {
                # drop file
                classValid <- FALSE
            } else {
                classValid <- TRUE
            }
```

write:
``` 
if (! all(myVclasses %in% classVknown)) {
                # skip file
                myNotes <- c(paste(i, " - was skipped because of unrecognized string in variant class column."))
                next
            }
```

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+    filesFinished <- 0
+    numFilesInvalid <- 0
+    for (i in fMAF) {
+
+        if (!silent) {
+            .pBar(i, length(fMAF))
+        }
+
+        if (fileValidity(i) & mafHeaderCheck(i) & snvValidity) {
+            hugo_ <- read.table(i, header = TRUE, fill = NA, stringsAsFactors = FALSE,
```
hyginn: Hm.
I would use ```readr::read.delim()``` and read exactly the columns you want, and none other, all at once. One read statement does it all.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+                # there was an invalid value
+                # report error in file and log
+                # drop file
+                # go to next
+                myNotes <- c(paste(i, " - File was dropped due to invalid attributes in column chr/class/type"))
+                numFilesInvalid <- numFilesInvalid + 1
+            } else {
+                # all tests passed
+                # add data to rSNV
+                fileasDT <- cbind(hugo_ , chr_ , start_, end_, strand_, var_class_,
```
hyginn: Unnecessary if you read it all at once...

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+                                  var_type_, ref_allele_, tum_all1_, tum_all2_,
+                                  tum_samp_barcode_, tum_sam_uuid_)
+                if (na.rm) {
+                    na.omit(fileasDT)
```
hyginn: ```na.omit``` is a function in the data.table package, isn't it? Doesn't it *return* the modified dataframe?
```myDF <- na.omit(mDF)```
Finally: don't you need to be selective about which columns trigger the omission? E.g not all datastes might have UUIDs to begin with ...

File: R/importSNV.TCGA.R [hunk 21]
```diff
@@ -0,0 +1,323 @@
...
+        # ==== WRITE LOG ===========================================================
+
+        if(writeLog) {
+
+            myTitle <- "importSNV.TCGA"
+
+            # Compile function call record
+            myCall <- character()
+            myCall[1] <- "importSNV.TCGA("
+            myCall[2] <- sprintf("fMAF = \"%s\", ", paste(fMAF, collapse=" "))
```
hyginn: That won't give the result you want if fMAF is a vector.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+            myCall[3] <- sprintf("rSNV = \"%s\", ", rSNV)
+            myCall[3] <- sprintf("silent = %s, ", as.character(silent))
+            myCall[4] <- sprintf("writeLog = %s)", as.character(writeLog))
+            myCall[5] <- sprintf("na.rm = %s)", as.character(na.rm))
+            myCall <- paste0(myCall, collapse = "")
```
hyginn: Missing the closing bracket of the call. Arguments not separated. Did you try this?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+            # indicate input object name(s)
+            #   (NA for this importSNV.TCGA())
+
+            # Record progress information
+            myNotes <- c(myNotes, sprintf("Files processed - ", filesFinished))
+            myNotes <- c(myNotes, sprintf("Total number of files in input fMAF - ", length(fMAF)))
+            myNotes <- c(myNotes, sprintf("Files dropped due to incomptibility - "), numFilesInvalid)
+
+            # indicate output object name(s)
+            myOutput = c("rSNV")
```
hyginn: That's not an output for which we can process metadata, UUID etc. In the way that we have designed the log-system, this event doesn't actually *have* input and output, because these are files that can be modified on disk. We would only attach metadata to R objects, eg. if we save this as an RDS. Which we don't, in this case.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,323 @@
...
+
+            # send info to log file
+            logEvent(eventTitle = myTitle,
+                      eventCall = myCall,
+                      #input = character(),
+                      notes = myNotes,
+                      output = myOutput)
+        }
+        return(rSNV)
```
hyginn: N/A - see my comment in the docstring.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,181 @@
...
+logFileName(fPath = tempdir(), setOption = TRUE)  # make tempdir() the log dir
+NL <- .PlatformLineBreak()
+# ==== END SETUP AND PREPARE ===================================================
+
+#context("Tools for importing MAF files into a rSNV file")
+options(stringsAsFactors=FALSE)
+
+test_that("parameter errors are correctly handled", {
+    # fMAF is not provided
+    expect_error(importSNV.TCGA())
```
hyginn: ```expect_error()``` uses a regex as a second argument, better use that. Otherwise you can't tell if the error you get is the error you expect - eg. whether it comes from the system, or from ```.checkArgs()```

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,181 @@
...
+    input2 <- c("PTCHD2" , 57540 , "WUGSC" , "GRCh38" , "chr1" , 11516130 , 11516130 , "+" , "Missense_Mutation" , "SNP" , "C" , "C" , "T" , "novel" , "TCGA-QQ-A8VF-01A-11D-A37C-09" , "TCGA-QQ-A8VF-10A-01D-A37F-09" , "Somatic" , "Illumina" , "HiSeq" , "2000" , "2109690d-e8b1-4b5c-9a39-21af273324cf" , "fe2daf93-56f1-4236-9b58-bf18f0e47d3a" , "c.1718C>T" , "p.Ala573Val" , "p.A573V" , "ENST00000294484")
+    input3 <- c("VWA5B1" , 127731 , "WUGSC" , "GRCh38" , "chr1" , 20319393 , 20319393 , "+" , "Missense_Mutation" , "SNP" , "C" , "C" , "T" , "novel" , "TCGA-QQ-A8VF-01A-11D-A37C-09", "TCGA-QQ-A8VF-10A-01D-A37F-09" , "Somatic" , "Illumina" , "HiSeq" , "2000" , "2109690d-e8b1-4b5c-9a39-21af273324cf" , "fe2daf93-56f1-4236-9b58-bf18f0e47d3a" , "c.853C>T" , "p.Pro285Ser" , "p.P285S" , "ENST00000375079")
+
+    trueIn <- rbind(input0,input1,input2,input3)
+    write.table(trueIn,  file="testIn", sep="\t", append=FALSE, row.names =
```
hyginn: Where is ```testIn```? Better do something like:
```
testIn <- tempfile()
... do whatever testing ...
expect_true(file.remove(testIn))
```

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,181 @@
...
+                    FALSE, col.names = FALSE, quote = FALSE)
+
+    # MAKE SMALL OUTPUT (trueOut) --------------------------------------------------
+    header <- c("sym", "chr", "start", "end", "strand", "class", "type", "aRef", "a1", "a2", "TCGA", "UUID")
```
hyginn: maybe more clear to write:
```
trueOut    <- c("sym", "chr", ...
trueOut[2] <- c("TMEM201", "chr1", 9607661, ...
trueOut[3] <- c("PTCHD2", "chr1",...
trueOut[4] <- c("VWA5B1", "chr1", ...
```
no intermediate assignment and ```rbind()``` required.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,181 @@
...
+
+    trueOut <- rbind(header,sample1,sample2,sample3)
+
+    # run the module on only fMAF as parameter (moduleOut)
+    moduleOut <- importSNV.TCGA(c("testIn"))
+    # check if manual output = module output
+    module <- read.csv(moduleOut, sep = "\t")
+
+    # TEST fMAF parameter ==========================================================
+    all.equal(trueOut, module)
```
hyginn: Hm. trueOut is a matrix, module is a data frame. That's already an issue. You haven't set stringsAsFactors = FALSE. So this will fail. Better to setup your Truth object as a data frame too.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,181 @@
...
+                    FALSE, col.names = FALSE, quote = FALSE)
+
+    # test rSNV input file
+    header <- c("sym", "chr", "start", "end", "strand", "class", "type", "aRef", "a1", "a2", "TCGA", "UUID")
+    snvInput1 <-c("TMEM201", "chr1", 9607661, 9607661, "+", "Missense_Mutation", "SNP", "C", "C", "T", "TCGA-QQ-A8VF-01A-11D-A37C-09", "2109690d-e8b1-4b5c-9a39-21af273324cf")
+    snvInput2 <- c("PTCHD2", "chr1", 11516130, 11516130, "+", "Missense_Mutation", "SNP", "C", "C", "T", "TCGA.QQ.A8VF.01A.11D.A37C.09", "2109690d-e8b1-4b5c-9a39-21af273324cf")
+
+    module <- rbind(header,snvInput1, snvInput2)
+
+    write.table(module,  file="module2", sep="\t", append=
```
hyginn: Write only to tempfiles! And remove them after testing.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,181 @@
...
+    # if rSNV is provided, check if number of columns in rSNV is worng
+    expect_error(importSNV.TCGA(c("testIn")), "testOutINVALIDcolumns")
+    # if rSNV is provided, check when header labels are wrong
+    expect_error(importSNV.TCGA(c("testIn")), "testOutINVALIDheaders")
+})
+
+test_that("silent and write works as intended", {
+    # if silent=TRUE, check if output to console is supressed
+    testF <- tempfile()
+    capture.output(importSNV.TCGA(c("testIn"), silent = TRUE), file = testF)
```
hyginn: no ```c()``` needed for a single item. In R, scalars *are* vectors (of length 1).

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,181 @@
+# testImportSNV.TCGA.R
```
hyginn: In all of your function calls, turn logging off all the time, except in one block where you check the log.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,181 @@
...
+    # run the module on fMAF, rSNV, silent, writeLog and na.rm as parameters
+    moduleOutNA <- importSNV.TCGA(c("testInNA"), silent=FALSE, writeLog=FALSE, na.rm=TRUE)
+    moduleNA <- read.csv(moduleOutNA, sep = "\t")
+    # add to prexisting rSNV file
+    all.equal(trueOutNA, moduleNA)
+    #===============================================================================
+})
+
+
+test_that("a corrupt input does not lead to corrupted output", {
```
hyginn: I think a check for an incomplete last line would be great (no carriage return). That's the hallmark of a truncated txt file. Probably whatever function you use to read the file in the first place throws an error or warning anyway that you can test for.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+#' @section Read MAF file: maf file is read by extracting require columns from maf file followed
+#' by checking if every item in the class, type and chr columns are from a known and required list of values.
+#' These columns as a data table are then added to an rSNV file.
+#'
+#' @section Write Log: log section writes file processing progress along with other errors
+#' that may occur.
+#'
+#' @param fMAF vector of MAF file names.
+#' @param rSNV rSNV file to merged new MAF data with.
+#' @param na.rm Remove all rows with missing or "NA" values and drop them from data frame.
```
hyginn: NA not in quotation marks. This is an expression of the R language, not a string.

R/importSNV.TCGA.R
hyginn: Check for missing values before you use them as parameters to .checkArgs()

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+#' @param silent Controls whether output to console should be suppressed. FALSE
+#'   by default.
+#' @param writeLog Controls whether writing the result to the global logfile is
+#'   enabled. TRUE by default.
+#'
```
hyginn: My comment about @return did not mean you should remove it. You should specify that nothing is returned.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+    # # Read header and one contents line
+    # tmp <- readLines(fName, n = 2)
+
+    # ==== VALIDATIONS =========================================================
+
+    # General parameter checks
+    cR <- character()
+    cR <- c(cR, .checkArgs(fMAF,         like = rep("FILE_E", length(fMAF)),
+                                                             checkSize = TRUE))
+    cR <- c(cR, .checkArgs(silent,       like = logical(1), checkSize = TRUE))
```
hyginn: You can check whether rSNV has a path prefix with the function ```dirname()``` which should be an existing directory.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+    # NL <- .PlatformLineBreak()
+    # # Read header and one contents line
+    # tmp <- readLines(fName, n = 2)
+
+    # ==== VALIDATIONS =========================================================
+
+    # General parameter checks
+    cR <- character()
+    cR <- c(cR, .checkArgs(fMAF,         like = rep("FILE_E", length(fMAF)),
+                                                             checkSize = TRUE))
```
hyginn: Check whether rSNV has a path prefix with the function ```dirname()``` which should be an existing directory.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+    cR <- c(cR, .checkArgs(silent,       like = logical(1), checkSize = TRUE))
+    cR <- c(cR, .checkArgs(writeLog,     like = logical(1), checkSize = TRUE))
+    cR <- c(cR, .checkArgs(na.rm,        like = logical(1), checkSize = TRUE))
+
+    if(length(cR) > 0) {
+        stop(cR)
+    }
+
+    # === Validate silent, writeLog, na.rm =================================
```
hyginn: ??? Pruthvi - all of these checks are being performed by ```.checkArgs()```  ... why?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+    # ==== Validata fMAF and rSNV  =======================================
+    if (missing(fMAF)) {
+        stop("missing parameter fMAF with no default: list of MAF file names")
+    } else if (is.null(fMAF)) {
+        stop("parameter fMAF cannot take value \"NULL\": should be a vector/string of MAF file name(s)")
+    } else if (is.numeric(fMAF)) {
+        stop("parameter fMAF cannot be numeric: should be a vector/string of MAF file name(s)")
+    }
+
+    if (missing(rSNV)) {
```
hyginn: I have no idea what you are doing here. Can you put in a comment about what problem this piece of code is trying to solve?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+        myNotes <- c(myNotes, paste("Number of rows originally in rSNV - ", rowsBefore))
+        snvValidity <- .rSNVheadercheck(rSNV)
+    }
+
+    # ==== READ DATA ===========================================================
+    filesFinished <- 0
+    numFilesInvalid <- 0
+
+    for (mafFile in fMAF) {
+        if (length(.fileValidity(mafFile)) + length(.mafHeaderCheck(mafFile)) + length(snvValidity) == 0) {
```
hyginn: Why are you checking for length(snvValidity) in the maf file loop?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+                     na.strings = " ", nrows=1)[,c(1,2,3,4,5,6,7,8,9,10,11,12)]
+        == validSNVheader)) {
+        headValid <- c(headValid, "Header of input rSNV file did not match the requirements.")
+    }
+    return(headValid)
+}
+
+.fileValidity <- function(file) {
+    validExt <- character()
+    if (! (strsplit(file, "\\.")[[1]][-1] %in% c("maf", "txt", "tsv"))) {
```
hyginn: This will fail if there is more than one period in the filename - which is a perfectly legal name. You need to extract only the extension. And don't use "file" when you mean the filename, and ```file()``` is a function anyway and is therefor not a good choice of variable name. 
```
regmatches(fName, regexpr("[^.]+$", fName))
```

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    if (missing(rSNV)) {
+        rSNVu <- .makeRandomString()
+
+        write("sym\tchr\tstart\tend\tstrand\tclass\ttype\taRef\ta1\ta2\tTCGA\tUUID",
+              rSNVu, append=FALSE)
+
+        snvValidity <- character()
+    } else if (is.null(rSNV)) {
```
hyginn: Again - this is stuff that ```.checkArgs()``` should handle.

rete_2017-03-23.1.log
```diff
@@ -0,0 +1,390 @@
+event | title  | importSNV.TCGA
```
wonjunetai: Should logs be removed or put into a `.gitignore`?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+            }
+
+            if (! classValid && typeValid && chrValid) {
+                # there was an invalid value
+                # drop file, go to next
+                numFilesInvalid <- numFilesInvalid + 1
+            } else {
+                # all tests passed
+                # add data to rSNV
+                if (na.rm) {
```
hyginn: Consider this bit of code (essentially from the examples of ```na.omit()```):
```
(myDF <- data.frame(x = c(1, 2, 3), y = c(0, 10, NA)))
na.omit(myDF[, 1])
na.omit(myDF[, 2])
```
... I don't think the result of your expression is what you intend it to be.

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+
+
+        # ==== WRITE LOG ===========================================================
+
+        if(writeLog) {
+
+            myTitle <- "importSNV.TCGA"
+
+            # Compile function call record
+            if (missing(rSNV)) {
```
hyginn: As above ... should not be missing.

rete_2017-03-23.1.log
hyginn: Delete this file.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,176 @@
...
+    expert_error(importSNV.TCGA(fMAF = testMAF,writeLog = c(TRUE, TRUE)), "write.log cannot have more than 1 values : should be logical TRUE/FALSE")
+
+    # if na.rm is not given logical input
+    expect_error(importSNV.TCGA(fMAF = testMAF,na.rm = 1), "na.rm cannot be numeric : should be logical TRUE/FALSE")
+    expert_error(importSNV.TCGA(fMAF = testMAF,na.rm = NULL), "na.rm cannot be NULL : should be logical TRUE/FALSE")
+    expert_error(importSNV.TCGA(fMAF = testMAF,na.rm = c(TRUE, TRUE)), "na.rm cannot have more than 1 values : should be logical TRUE/FALSE")
+
+})
+
+test_that("a sane input gives an expected output with differnt input parameters", {
```
hyginn: *different

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,176 @@
...
+
+    # MAKE SMALL INPUT (testIn) ----------------------------------------------------
+    testIn <- tempfile()
+    input0 <- c("Hugo_Symbol" , "Entrez_Gene_Id" , "Center" , "NCBI_Build" , "Chromosome" , "Start_Position" , "End_Position" , "Strand" , "Variant_Classification" , "Variant_Type" , "Reference_Allele" , "Tumor_Seq_Allele1" , "Tumor_Seq_Allele2" , "dbSNP_RS" , "dbSNP_Val_Status" , "Tumor_Sample_Barcode" , "Matched_Norm_Sample_Barcode" , "Match_Norm_Seq_Allele1" , "Match_Norm_Seq_Allele2" , "Tumor_Validation_Allele1" , "Tumor_Validation_Allele2" , "Match_Norm_Validation_Allele1" , "Match_Norm_Validation_Allele2" , "Verification_Status" , "Validation_Status" , "Mutation_Status" )
```
hyginn: :-(

```
trueIn <- character()
trueIn[1] <- ....
trueIn[2] <- ...
etc.
```

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,176 @@
...
+    input2 <- c("PTCHD2" , 57540 , "WUGSC" , "GRCh38" , "chr1" , 11516130 , 11516130 , "+" , "Missense_Mutation" , "SNP" , "C" , "C" , "T" , "novel" , "TCGA-QQ-A8VF-01A-11D-A37C-09" , "TCGA-QQ-A8VF-10A-01D-A37F-09" , "Somatic" , "Illumina" , "HiSeq" , "2000" , "2109690d-e8b1-4b5c-9a39-21af273324cf" , "fe2daf93-56f1-4236-9b58-bf18f0e47d3a" , "c.1718C>T" , "p.Ala573Val" , "p.A573V" , "ENST00000294484")
+    input3 <- c("VWA5B1" , 127731 , "WUGSC" , "GRCh38" , "chr1" , 20319393 , 20319393 , "+" , "Missense_Mutation" , "SNP" , "C" , "C" , "T" , "novel" , "TCGA-QQ-A8VF-01A-11D-A37C-09", "TCGA-QQ-A8VF-10A-01D-A37F-09" , "Somatic" , "Illumina" , "HiSeq" , "2000" , "2109690d-e8b1-4b5c-9a39-21af273324cf" , "fe2daf93-56f1-4236-9b58-bf18f0e47d3a" , "c.853C>T" , "p.Pro285Ser" , "p.P285S" , "ENST00000375079")
+
+    trueIn <- rbind(input0,input1,input2,input3)
+    write.table(trueIn,  file=testIn, sep="\t", append=FALSE, row.names =
+                    FALSE, col.names = FALSE, quote = FALSE)
+
+    # MAKE SMALL OUTPUT (trueOut) --------------------------------------------------
+    header <- c("sym", "chr", "start", "end", "strand", "class", "type", "aRef", "a1", "a2", "TCGA", "UUID")
```
hyginn: same as above ...

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,176 @@
...
+
+    trueOut <- rbind(header,sample1,sample2,sample3)
+
+    # run the module on only fMAF as parameter (moduleOut)
+    moduleOut <- importSNV.TCGA(testIn)
+    # check if manual output = module output
+    module <- read.table(moduleOut, sep = "\t", stringsAsFactors = FALSE)
+
+    # TEST fMAF parameter ==========================================================
+    all.equal(trueOut, module)
```
hyginn: ```all.equal()``` will either return TRUE or a character vector describing the differences. But this won't interface with testthat. What you need is ```expect_equal()```.

tests/testthat/testImportSNV.TCGA.R
```diff
@@ -0,0 +1,176 @@
...
+
+test_that("a corrupt input does not lead to corrupted output", {
+    # if fMAF file open does not follow the 2.4 or agreed upon version either STOP or DROP file
+    # expect_error(importSNV.TCGA(c("testIn2.4.1")))
+
+    # if fMAF file open does not have the required columns at the
+    #           right column number STOP or DROP file
+})
+
+test_that("silent and writeLog works as intended", {
```
hyginn: You are mentioning ```writeLog``` here but not testing for it ...

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+        snvValidity <- character()
+    } else if (is.null(rSNV)) {
+        stop("parameter rSNV cannot take value \"NULL\": should be absent or existing rSNV file name")
+    } else if (is.numeric(rSNV)) {
+        stop("parameter rSNV cannot be numeric: should be absent or existing rSNV file name")
+    } else if (length(rSNV) > 1) {
+        stop("parameter rSNV cannot take more than one file as input")
+    } else {
+        rSNVu <- rSNV
+        rSNVf <- read.table(rSNV, header = TRUE)[,1]
```
hyginn: This are potentially very large files. This function should not read an entire file into memory, just to count the number of lines! There are several ways to do this, most direct is probably
```
system(sprintf("wc -l %s", rSNV))
```
and parsing the number from the output. I believe this also works on Windows...

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+    # ==== READ DATA ===========================================================
+    filesFinished <- 0
+    numFilesInvalid <- 0
+
+    for (mafFile in fMAF) {
+        if (length(.fileValidity(mafFile)) + length(.mafHeaderCheck(mafFile)) + length(snvValidity) == 0) {
+
+            fileasDT <- read.table(mafFile, header = TRUE, fill = NA,  stringsAsFactors = FALSE,
+                                        na.strings = " ")[,c(1,5,6,7,8,9,10,11,12,13,16,33)]
+
```
hyginn: so what's wrong with ```readr::read.delim()```?

R/importSNV.TCGA.R
```diff
@@ -0,0 +1,268 @@
...
+        # ==== WRITE LOG ===========================================================
+
+        if(writeLog) {
+
+            myTitle <- "importSNV.TCGA"
+
+            # Compile function call record
+            if (missing(rSNV)) {
+                myCall[1] <- "importSNV.TCGA("
+                myCall[2] <- sprintf("fMAF = \"%s\", ", fMAF)
```
hyginn: That still won't give you the result you want if fMAF is a vector. Try it.

R/importSNV.TCGA.R
Pruthv1: Yeah, I thought so too. But from one of your earlier comments about using the second parameter of expect_error() which is a regex. I thought you wanted me to throw my own errors from this function and check for them separately.
Fixed now.

PR #22, opened 2017-03-21
Combine function 

TonkatsuD:
Pseudo-code
testthat design

### DIFF
--- a/R/combineSNV_CNA.R
+++ b/R/combineSNV_CNA.R
@@ -20,42 +20,45 @@
 #'
 #'
 #' @examples
-#' \dontrun{combine(fName=file_names, fgX="combined_data.txt", silent=TRUE, noLog=FALSE)}
+#' \dontrun{combine(fName=file_names, fgX="combined_data.RDS", silent=TRUE, noLog=FALSE)}
 #'
 #' @export
 combineSNV_CNA <- function(fname, fgX, silent=FALSE, noLog=FALSE) {
     #' pseudo code
-    #' hashVector <- ()
-    #' index <- 1
+    #' hashTable <- new.env()
     #' read from input files and store hash
     #' for (i <- 1, i <= length(fname), i++) {
     #'  fcurrent <- fname[i]
     #'  if fcurrent is CNA {
     #'      loadRDS(fcurrent)
     #'      for each variation:
-    #'          cnaCount <- count copy number variation
-    #'          hashVector[index] <- hash(cnaCount)
-    #'          index++
+    #'          key = hash(<gene symbol>:CNA:<as.character(round(<copy number>))>)
+    #'          if key is in hashTable:
+    #'              hashTable[[key]] <- hashTable[[key]] + 1
+    #'          else:
+    #'              hashTable[[key]] <- 1
     #'  }
     #'  else if fcurrent is MUT {
     #'      load(fcurrent)
     #'      for each row:
-    #'          mutCout <- count observed mutation
-    #'          hashVector[index] <- hash(mutCount)
-    #'          index++
-    #'  }
+    #'          key = hash(<gene symbol>:SNV:<position>:<variant class>)
+    #'          if key is in hashTable:
+    #'              hashTable[[key]] <- hashTable[[key]] + 1
+    #'          else:
+    #'              hashTable[[key]] <- 1
     #' }
-    #' sort hash
-    #' hashVector <- sort(hash_vector)
-    #' count <- (length(hash_vector))
     #'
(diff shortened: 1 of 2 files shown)


R/combineSNV_CNA.R
```diff
@@ -0,0 +1,68 @@
+# combineSNV_CNA.R
+
+#' Combine mutation data from CNA and SNV
+#'
+#' \code{combineSNV_CNA} loads the vector of preprossed dataset from filtered CNA or SNV files,
```
hyginn: *preprocessed

R/combineSNV_CNA.R
```diff
@@ -0,0 +1,68 @@
...
+#'  creates hash keys corresponding to the input file type, combine and save the hashed
+#'  keys to a gX format file as output file.
+#'
+#' Details.
+#' @section gX files:
+#'  gX files contain data of SNV and CNA combined for each gene as one object, sorted per
+#'  gene. Tab-delimited text file with header.
```
hyginn: I've just changed the spec to make this a data frame RDS. Makes everyone's life easier.

R/combineSNV_CNA.R
```diff
@@ -0,0 +1,68 @@
...
+#'
+#'
+#' @param fName A vector of path of fully qualified file names of SNV and/or CNA files.
+#' @param fgX The fully qualified filename/path of a gX output file.
+#' @param silent Boolean option for writing combiling process information to console.
+#' @param noLog Boolean option for log results.
```
hyginn: No. Consistent with all other functions now this is ```writeLog = TRUE```

R/combineSNV_CNA.R
```diff
@@ -0,0 +1,68 @@
...
+#' @return number of hashed keys stored to the gX output file.
```
hyginn: I would not return a value that is not requested by the downstream function. Save file and log output as side-effects should be sufficient.

R/combineSNV_CNA.R
```diff
@@ -0,0 +1,68 @@
...
+#'
+#'
+#' @examples
+#' \dontrun{combine(fName=file_names, fgX="combined_data.RDS", silent=TRUE, noLog=FALSE)}
+#'
+#' @export
+combineSNV_CNA <- function(fname, fgX, silent=FALSE, noLog=FALSE) {
+    #' pseudo code
+    #' hashTable <- new.env()
```
hyginn: Use the hash package. You'll need support to fetch all keys, so you can sort them to process output.

R/combineSNV_CNA.R
```diff
@@ -0,0 +1,68 @@
...
+    #'              hashTable[[key]] <- 1
+    #' }
+    #'
+    #'
+    #'
+    #' retrieve all keys from hashTable
+    #' hashVector = hashTable[!is.na(hashTable)] or hashVector <- keys(hashTable)
+    #' sort hashVector
+    #' strplit them to sym, type, pos and class
+    #' store retrieve 4 classes in a dataframe
```
hyginn: Careful: you'll need to create appropriate NAs for the CNAs for column ```$pos```

tests/testthat/testCombine.R
```diff
@@ -0,0 +1,34 @@
...
+    # test if returned number is correct
+    # test if a log file is saved if log is True
+    # test if process are sent to console is silent is True
+    # test if output file fgX is saved under the path
+
+})
+
+
+test_that("combine received corrupted input data file", {
+    # try both empty SNV and CNA datafile, expect empty fgX dataframe
```
hyginn: Not sure - here I would only test mismatch between headers and data, and incomplete lines (incomplete file). I don't think the combine code requires requires a full data validation - if so, that should be done in the actual import routines.

tests/testthat/testCombine.R
```diff
@@ -0,0 +1,34 @@
+# testCombine.R
```
hyginn: - use the setup/teardown headers from ```tests/testthat/testDummy.R```
- always turn logging off, except for one block in which you explicitly test logging
- use ```tempdir()``` or ```tempfile()``` for intermediate files to be created, and delete them after testing.

R/combineSNV_CNA.R
```diff
@@ -0,0 +1,68 @@
...
+    #' for key in hashVector:
+    #'  retrieve value from hashTable using keys in hashVector
+    #' count <- length(hashVector)
```
hyginn: Don't forget adding metadata

hyginn:
Looks good in principle. On with the test design...

PR #23, opened 2017-03-21
edgeThresh initial submission 

FarzanT:


R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#' choosing 2 edges and swapping their target endpoints (i.e. B in the
+#' edge A → B). Finds the value δ (edge weight) for each random graph such that
+#' the maximum size of any component in the graph is Lmax, and returns the
+#' median of N δ values.
+#'
+#' @section Swapping Edges:
+#'     Bidirectional edges must only be swapped with each other, therefore the
+#'     swapping operation will first be applied to bidirectional edges,
+#'     Q * |E_b| times, where E_b is a bidirectional edge. Since all 4 incident
+#'     nodes in 2 bidirectional edges are receiving endpoints, one node will be
```
hyginn: As written, this is not a correct description of an edge swapping operation.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#'     randomly selected from each edge, and a new edge is formed between the
+#'     selected as well as the remaining vertices.
+#'     Then Q * |E_u| unidirectional edge swaps take place.
+#'     At each unidirectional edge swap, only the receiving endpoints are
```
hyginn: Correct: you are preserving indegree and outdegree.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#'     swapped e.g. (A → B, C → D) to (A → D, C → B).
+#'     In both cases, a swap should NOT result in:
+#'     Multiple edges: Possibly result of a bidirectional edge swap.
```
hyginn: *multiple

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#'     Bidirectional edges must only be swapped with each other, therefore the
+#'     swapping operation will first be applied to bidirectional edges,
+#'     Q * |E_b| times, where E_b is a bidirectional edge. Since all 4 incident
+#'     nodes in 2 bidirectional edges are receiving endpoints, one node will be
+#'     randomly selected from each edge, and a new edge is formed between the
+#'     selected as well as the remaining vertices.
+#'     Then Q * |E_u| unidirectional edge swaps take place.
+#'     At each unidirectional edge swap, only the receiving endpoints are
+#'     swapped e.g. (A → B, C → D) to (A → D, C → B).
+#'     In both cases, a swap should NOT result in:
```
hyginn: Under what conditions **could** a swap result in multiple edges? You need to define the condition so you can test for it.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#'     Multiple edges: Possibly result of a bidirectional edge swap.
+#'     Self-edges: e.g. (A ↔ B, C ↔ A) to (A ↔ A, B ↔ C)
+#'     Disconnected components: Considering (A → B, C → D), if A and B are
```
hyginn: Assuming you start with a connected component, what are the conditions that could result in a disconnected component? Can you test for that efficiently? Also note that the components need to be *strongly* connected. Tarjan's algorithm is linear *O*(|V| + |E|), I hope that's what igraph uses. Nevertheless: Q* |E| * Tarjan's makes this overall *O*(|E|^2). Needs some thought.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#'     in the same component, they should still be in the same component when
+#'     (A → D, C → B) This guarantees that the graph is still connected if
+#'     the starting graph was also connected.
+#'     ### What about (A → B, A → D) to (A → D, A → B)? Will this be
```
hyginn: This swap does not conserve the node degree of B and D.

If you simply ```sample()``` without replacement from appropriate node lists, that problem won't occur.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#'     cosidered a swap? (counted towards Q * |E|) ###
+#'
+#' @section Maintaining Connected Components:
```
hyginn: Since the target outcome is to measure order of the largest connected component, I would use only the largest connected component of the input for the computation in case the input has more than one component. Otherwise, a small component would trigger the threshold even without edge removal.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#'     In order to ensure nodes maintain their components during permutation,
+#'     the components of the graph is measured initially, and the component
+#'     each node belongs to is stored in a separate data structure (hash table?).
+#'     After each swap, the new component of the sending endpoint (A in A → B)
+#'     is the same as the component the receiving endpoint belongs to. Thus
+#'     the component maintenance is acieved in 2 steps (one query and one
+#'     comparison) for each receiving node, and thus 4 * Q * |E| steps are
+#'     required.
+#'
+#'     ## Why not use a binary search style approach to find the minimum delta?
```
hyginn: Indeed, why not? That sounds like an improvement **IF** you can identify a reasonable range. Note that we haven't said anything about increments anyway ...

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#'     ## Can calculation of delta be parallelized?
```
hyginn: Yes: one can trivially parallelize over the N networks, and over the *ord* levels of each network. These are independent computations. But adding parallelization **now** is premature..

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#' @section Incrementing δ and Lmax: ## How much should delta be incremented for efficieny as well as accuracy?##
+#'     After creating the permutated graph, the following procedure is run on
+#'     the graph for each Lmax starting from 2 to the Lmax given as an argument.
+#'     On each such run, and a small value for δ is chosen (0.01) , and
+#'     the strongly connected component discovery is done. If the size of the
+#'     biggest component after such run is smaller than Lmax, increment δ by
+#'     (0.01), and redo component disovery. As soon as the largest component has
+#'     size Lmax (for current run), δ is recorded, and Lmax is incremented by 1.
+#'     Once current Lmax is the same as the given Lmax, run the procedure on the
+#'     next graph.
```
hyginn: Your suggestion for a binary search is good - perhaps there are even better methods, essentially this is a root finding problem on a function that does not have continuous derivatives.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+#' @param Q	The number of edge-swapping permutations to produce one of the
+#' permuted networks is Q * |E|.
+#' @param silent Boolean. Default: FALSE. Whether or not to write
+#' progress information to console.
+#' @param writeLog Boolean. Default: TRUE Whether or not to log results.
+#'
+#' @return The Equilibrated Gene Graph EGG with δ values as additional
+#' graph metadata.
+#'
+#' @author Farzan Taj farzantaj@outlook.com
```
hyginn: No separate author identification in collaborative code.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+                   N = 100,
+                   Q = 100,
+                   silent = FALSE,
+                   writeLog = TRUE) {
+    # NOTE: Most comments, only as a confirmation/question
+
+
+    # Check arguments for validity, generate error if necessary and stop
+    # Read EGG
+    # Use the helper function ".permute" to create permutations,
```
hyginn: Why do you want to put N graphs into memory all at the same time if you are working only on one at a time?

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+    # which takes a copy of EGG, and N, and returns N random
+    # graphs as a list (?) of objects
+
+
+    # If variables are assigned in the local environmet, the helper functions
```
hyginn: Generally not a good idea. Sometimes necessary, probably not here.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+    # can use them.
+    # Generate the list of edges, vertices and vertex count available in EGG
```
hyginn: ???? Just pass an EGG graph to the helper ... it's going to be only read-from anyway.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+    allEdges <- igraph::get.edgelist(EGG)
+    allVertices <- igraph::V(EGG)
+    numVertices <- igraph::vcount(EGG)
+
+    # Call permute on EGG, store returned list of graphs as an object
+    graphList <- .permute(EGG, N)
```
hyginn: call this ```permuteGraph()```

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+    masterDeltaList <- vector("list", length = N)
+
+    # For each graph:
+
+    # Create a vector storing min δ ord for each maxComponentSize
```
hyginn: This vector needs to move outside the loop. Many δs are computed for each graph.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+    # (= is used to only store objects in current loop)
+    graphDeltas = vector("numeric", Lmax - 1)
+    currentDelta = 0.01
+    currentGraph = graphList[i]
+    currentEdges = igraph::E(currentGraph)
+
+    # Set maxComponentSize to 2 initially, and increase to a max of Lmax.
+    maxComponentSize = 2
+
+    # For each Lmax, set δ to 0.01 (?), remove all edges that have weights that
```
hyginn: Needs to work on a copy of the current graph.

hyginn: I think one probably should write a .nextDelta() function here that takes care of defining the best next value. Initially, just work from increments/linear search, but you can then drop in a smarter replacement with binary search or something.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+    # are below such δ.
+    currentGraph <- igraph::delete.edges(currentGraph, which(currentEdges$weight <= currentDelta))
+
+    # Find the largest 'strongly' connected component:
+    largestComponent <- max(igraph::components(currentGraph, mode = "strong")$csize)
+
+    # If largest component has size less than Lmax,
```
hyginn: That does not implement the algorithm. The algorithm has to increase δ to make the largest connected component smaller, and record δ values for which the order of the component falls into the target range.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+            masterDeltaList[[i]] <- graphDeltas
+            break
+        }
+    }
+    # (What if the first δ leads to a size equal to maxComponentSize? Should δ
+    # be decreased instead? By how much should δ be incremented at each step?)
+
+    # Find the minimum δ for each graph at each maxComponentSize
+
+    # After generating N vectors of size (Lmax - 1), find the median δ for each
```
hyginn: See my note at 42

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+    # δ ord (maxComponentSize). i.e. Lmax - 1 median values.
+    deltaOrd <- vector("numeric", Lmax - 1)
+    for (i in 1:(Lmax - 1)) {
+        deltaOrd[i] <- median(sapply(masterDeltaList, FUN = "[", i))
+    }
+
+    # Attach the vector as graph metadata to the original EGG and return.
+    # (Is this the correct way?)
+    attributes(EGG) <- deltaOrd
```
hyginn: No.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+
+    # (Can this be parallelized? Since the processes are not interdependent)
```
hyginn: Too early to even think about this.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+
+    # Write to Console:
+    # Update the fraction of N graphs that are done with permutations, (can
+    # also update the rough percentage of Q * |E| swaps that are done, but can
+    # slow down the process). This is done in .permute.
```
hyginn: I don't understand what you mean here. Console output is a general overview of the state of the process, not an exact record of data - unless you are requesting user input. In general, some progress indicator is good if time to completion takes longer than about five seconds.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+
+#' Helper function to create N permutations of the EGG graph.
+#'
+#' @param EGG The given EGG graph from parent function.
+#' @param N The number of permuted graphs that must be generated.
+#' @param Q The number of edge-swapping permutations to produce one of the
+#' permuted networks is Q * |E|.
+#'
+#' @return List of N igraph objects, permutations of the given EGG graph
+.permute <- function(EGG, N, Q) {
```
hyginn: as above: input one graph, return one permuted graph.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+    # Calculate the component each vertex belongs to
```
hyginn: Assume incoming graph is a single component.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+    E <- t(apply(X = allEdges, MARGIN = 1, FUN = sort))
+    bidirectionalEdges <- allEdges[duplicated(E) | duplicated(E, fromLast = TRUE)]
+    numBidirectional <- (length(bidirectionalEdges)) / 2
+    # Since they are all ordered, we can pick only the first half, and group
+    # each pair
+
+    # Better to use a list of tuples to keep track of 'bidirectional' edge pairs
+    # as they are simply a pair of unidirectional edges
+    bidirPairList <- vector("list", length = (numBidirectional / 2))
+    for (i in seq(1, numBidirectional, 2)) {
```
hyginn: You don't need that. Bidirectionality is implied. Separate just one of each pair out into bidirected category, and whenever a swap is required, swap both.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+        bidirPairList[[(i %/% 2) + 1]] <- c(bidirectionalEdges[i], bidirectionalEdges[i + 1])
+    }
+
+    ### For each of the N permutated graphs:
+
+    # Create a new copy of the original EGG graph
+    currentGraph <- EGG
+    ## Swap bidirectional edges first:
```
hyginn: Why do you swap bidirected first? You can just swap randomly from both categories. This is a stochastic swap, not iterating over the edges.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+        # Use add_edges() and delete_edges()
+        # Can use igraph::any_multiple() to check if multiple edges are created
+        # after each swap, do not increment i if this is the case.
+
+        # Check the new components of all nodes against their previous ones,
+        # using the hash structure created above, discard change if components
+        # are changed (Does this address the disconnected components issue?)
+    }
+
+    # Bidirectional edge swap must not result in multiple edges, thus the new
```
hyginn: Correct.

R/edgeThresh.R
```diff
@@ -0,0 +1,284 @@
...
+
+    # igraph::keeping_degseq() allows rewiring while preserving "total" degree
+    # distribution (of the graph), not sure if it can be used here
+    # a <- rewire(graph = EGG, with = keeping_degseq())
+
+    # Function degree() with modes: "in", "out" and "all" can be used to check
+    # how many edges point to and from a node.
+
+
+    ### After both kinds of edge swap (or preparation), each node will be
```
hyginn: I don't think a pair swap can split a graph into weakly connected components. Not sure about strongly connected components though.

hyginn:
Comments added - let's move on to designing tests.

R/edgeThresh.R
FarzanT: I'm currently addressing other issues as well, but just to be sure with this one: Is it sufficient to find all articulation points and identify edges that connect to them (incoming and outgoing), and during a swap, an edge between two adjacent articulation points is not to be swapped with an incoming edge to another articulation point. Is this correct/enough/efficient? (I have tried this with a component, and it results in disconnected components. Not sure if this is can be the only case)

FarzanT:
I guess I have to close this pull request to be able to open another one, which is for unit testing.
Please answer my comment regarding disconnected components, as I am still uncertain.

R/edgeThresh.R
FarzanT: Never mind, I just discovered this old idea is known as bridges, and I'm simply going to find them in the EGG.

FarzanT:
@hyginn "Comments added - let's move on to designing tests."
Ok now I'm going back to addressing comments

hyginn:
Curious. I did write that. Now I remember that I thought - there's really quite a lot not clear about the algorithm. But the required behaviour is reasonably well defined. Input and output edges should be no more similar than random. Node degrees conserved. Output remains strongly connected. Number of bidirected edges conserved. Structure of the delta values in the metadata... So yes, you might as well define the tests. 

My irritated response came when I was reviewing the files in the pull request from the top down - but it seems you actually had meant for me to only review the tests.

FarzanT:
Sorry but regarding connectedness, I thought that the largest connected component should remain connected, not necessarily "strongly" connected. Is this not the case?

hyginn:
Would that make sense?

FarzanT:
@hyginn I just went back and had a look at the HotNet2's supplementary material, and they only mention that they ensure the network remains connected during the swaps. Then, while checking for the delta values, the size of the largest strongly connected component is considered.
So I guess we don't have to worry about strongly connected components remaining strongly connected during the permutations.

hyginn:
That was not what I asked.

hyginn:
Put this differently: what features of the input graph should the permutation conserve? Incidentally that is also going to define your tests ...

FarzanT:
Number of vertices, number of edges, node degrees, members of all components (the largest component in this case) i.e. they should stay in the same component. Is this not the case? This is what I understand from our design and the supplementary material of HotNet2

R/edgeThresh.R
FarzanT: @hyginn I need to be sure that node that disallowing swapping between bridges is the right way to prevent disconnected components, before I can move on. Unfortunately igraph does not have such a function yet.

PR #30, opened 2017-03-22
Score.R pseudocode.

jkitaygorodsky:
https://github.com/hyginn/rete/compare/master...jkitaygorodsky:master#diff-e068974d6182813df1bdba17f4a44f33


R/score.R
```diff
@@ -0,0 +1,57 @@
+# score.R
+
+#' Produce a gH object: a hash of genes and their associated scores
+#' (likelihood that observed mutations exceed background mutations at FDR of 1%)
+#'
+
+score <- function(gX, gH, silent = FALSE, noLog = FALSE) {
```
hyginn: The common part of all our function signatures is ```writeLog = TRUE```. I have even commented on that change explicitly on Riot.

R/score.R
```diff
@@ -0,0 +1,57 @@
...
+    
+    check parameters with .checkArgs()
+    
+    open gX file
+    
+    # for each gene, check if it already in the gH hash and if yes, do not calculate its score (in case have a repeat of a gene in gX, would just use first calculated score)
+    
+    for gene in gX {
+        if gene already in gH {
+            break
```
hyginn: this is **next** not **break**!

R/score.R
```diff
@@ -0,0 +1,57 @@
...
+#' (likelihood that observed mutations exceed background mutations at FDR of 1%)
+#'
+
+score <- function(gX, gH, silent = FALSE, noLog = FALSE) {
+    
+    check parameters with .checkArgs()
+    
+    open gX file
+    
+    # for each gene, check if it already in the gH hash and if yes, do not calculate its score (in case have a repeat of a gene in gX, would just use first calculated score)
```
hyginn: I don't agree with dropping data. And anyway: have you looked at how a gX is structured? I think you are misunderstanding this: every gene may have **many** records of changes included. They are sorted for convenience, but they don't really have to be. 

Also: gX contains SNV **and** CNV data. How will you handle the CNV data?

R/score.R
```diff
@@ -0,0 +1,57 @@
...
+    
+    for gene in gX {
```
hyginn: I would iterate through the entire data frame (it's an rds object, so you get the whole thing at once anyway), and collect the information as you go through it. So perhaps build your hash from the genes you find in ```unique(gX$sym)```, and then go through the frame row by row.

R/score.R
```diff
@@ -0,0 +1,57 @@
...
+        if gene already in gH {
+            break
+        } else {
+    
+            # define background mutation rate for gene
+            back_mut = (missense + nonsense)/silent   # would be more accurate to use the background mutation data from Samocha et al., 2013 (http://www.nature.com/ng/journal/v46/n9/full/ng.3050.html)
```
hyginn: A: In what sense is ```(missense + nonsense)/silent``` a "background" mutation rate? What are you calculating here?

B: Can you calculate a score according to Samocha *et al.* with the data you **have**? If so, how?

R/score.R
```diff
@@ -0,0 +1,57 @@
...
+            
+            # calculate observed mutation rate for gene
+            observed_mut = missense + nonsense
```
hyginn: That's a count, not a "rate".

R/score.R
```diff
@@ -0,0 +1,57 @@
...
+            
+            
+            # assume Poisson distribution (appropriate distribution, since mutation is rare event, so both the background and observed mutation rates are expected to be low)
+            # determine cut-off for lambda with FDR = 0.01
+            
+            lower_threshold <- qpois(0.01, back_mut, lower.tail = TRUE)
```
hyginn: qpois' lambda is "a vector of means". Is back_mut a vector? Of means?

Can you outline the process of your statistics more clearly here? I am not sure that this is correct. (I'm not saying it's not though.)

Working backwards:
- FDR is applied to a vector of p values (cf. ```p.adjust()```)
- This means you need a p value derived from the ratio of non-silent vs. silent mutations. (Incidentally, note that you need to take the number of sites into account).
- For that, you need an idea of the distribution of the ratios, which you can convert into a p-value of an observed excess. The distribution under the null-model is going to depend on the number of observed mutations. Few mutations will have large spread, many mutations are going to get nearly the same ratio every time.

Looking forward to your thoughts ...

R/score.R
```diff
@@ -0,0 +1,57 @@
...
+        } else {
+    
+            # define background mutation rate for gene
+            back_mut = (missense + nonsense)/silent   # would be more accurate to use the background mutation data from Samocha et al., 2013 (http://www.nature.com/ng/journal/v46/n9/full/ng.3050.html)
+            
+            # calculate observed mutation rate for gene
+            observed_mut = missense + nonsense
+            
+            
+            # assume Poisson distribution (appropriate distribution, since mutation is rare event, so both the background and observed mutation rates are expected to be low)
```
hyginn: Review your approach: see my comment below.

R/score.R
```diff
@@ -0,0 +1,57 @@
...
+            # determine likelihood that observed mutations exceed background mutations at an FDR of 0.01 (using the cut-off calculated above)
+            
+            score <- (dpois(observed_mut, cut_off)/dpois(back_mut, cut_off))
+            
+            # store gene and associated score in gH file
+            
+            gH[gene] <- score
+        }
+    
+    }
```
hyginn: - write metadata
- write log
- sumarize process to console
- save gH as .rds

R/score.R
```diff
@@ -0,0 +1,57 @@
+# score.R
+
+#' Produce a gH object: a hash of genes and their associated scores
+#' (likelihood that observed mutations exceed background mutations at FDR of 1%)
```
RicardoRam: include @param and @return in dunction documentation

RicardoRam:
Please include the R CMD check.

PR #35, opened 2017-03-23
testEdgeThresh.R (unit tests)

FarzanT:
This reverts commit 25f6cf017fda703c4987c7d1fe6f47ded8d67ea0.
*Added simple "parameter check" test cases + design and preliminary code of other test sections


tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+context("Graph permutation and delta calculation")
+
+
+# ==== BEGIN SETUP AND PREPARE =================================================
+# OLOG <- as.character(getOption("rete.logfile"))   # save original logfile name
+# logFileName(fPath = tempdir(), setOption = TRUE)  # make tempdir() the log dir
+# NL <- .PlatformLineBreak()
+# if (file.exists(logName)) { file.remove(logName)}
+
+# ((Is rete.EGGprototype only the barebones valid EGG, or does it contain testable
```
hyginn: The prototypes ultimately should contain all features that .checkArgs() would need to perform a validity check.

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+nVertices <- igraph::vcount(EGG)
+nEdges <- igraph::ecount(EGG)
+# ==== END SETUP AND PREPARE ===================================================
+
+
+test_that("edgeThresh parameter errors are correctly handled", {
+    # Try each parameter missing in turn (only EGG has no defaults)
+    expect_error(edgeThresh())
+
+    # Try parameters out of expected bounds
```
hyginn: Check error strings. (It would be helpful if you read comments to others' pull requests.)

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    expect_error(edgeThresh(myEGG, Lmax =  1))
+    expect_error(edgeThresh(myEGG, Lmax =  0))
+    expect_error(edgeThresh(myEGG, Lmax = -1))
+    expect_error(edgeThresh(myEGG, Lmax = nLargst - 1))
```
hyginn: What is this?

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    expect_error(.permuteGraph(graph = myEGG, Q = -1))
+
+    expect_error(.permuteGraph(graph = 10,    Q = 100))
+
+}
+# ((Should valid inputs different from rete.EGGprototype be used?))
+# ((The testing below can get quite computationally expensive...))
+test_that("a sane input to .permuteGraph gives an expected output", {
+
+    ### Test for Q = 1, 10, 20, 50, 100, 200
```
hyginn: What will this achieve?

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
+# edgeThresh.R
```
hyginn: You need to comment on *why* you are performing tests if they are not trivial.

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    expect_error(edgeThresh(myEGG, Lmax = nLargst - 1))
+
+    expect_error(edgeThresh(myEGG, Lmax = 20, N =  0))
+    expect_error(edgeThresh(myEGG, Lmax = 20, N = -1))
+
+    expect_error(edgeThresh(myEGG, Lmax = 20, Q =  0))
+    expect_error(edgeThresh(myEGG, Lmax = 20, Q = -1))
+
+
+    # Try NULL and length-zero parameters: be sure none lead to an erroneous
```
hyginn: While that's certainly true of input parameters, this consideration is more important for the code itself - however this can't be easily checked here because you are already assuming the parameters can't go out of range.

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+
+}
+# ((Should valid inputs different from rete.EGGprototype be used?))
+# ((The testing below can get quite computationally expensive...))
+test_that("a sane input to .permuteGraph gives an expected output", {
+
+    ### Test for Q = 1, 10, 20, 50, 100, 200
+    myPermutation <- .permuteGraph(myEGG, Q = 1)
+
+    # Check for self-loops and multiple edges
```
hyginn: What happens if your input graph is not simple?

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    expect_true(igraph::is.simple(myPermutation))
+
+    # Check that the size of the largest connected component has not changed
```
hyginn: Strongly connected!

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    expect_equal(max(igraph::components(myPermutation)$csize), nLargest)
+
+    # Check that overall size has not changed
+    expect_equal(igraph::vcount(myPermutation), nVertices)
+
```
hyginn: Missing: checks for degree, size, number of bidirected edges ...

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    expect_true(igraph::is.simple(myPermutation))
+    expect_equal(max(igraph::components(myPermutation)$csize), nLargest)
+    expect_equal(igraph::vcount(myPermutation), nVertices)
+
+})
+# ((Is it possible to pass permuted graphs from the previous test section
+# to the next one, if they pass? This would minimize a lot of computation))
+test_that("a sane input to edgeThresh gives an expected output", {
+    # Provide small input data, provide small output data.
```
hyginn: I don't see why you are benchmarking your results by varying Q in this test. You really need to explain to me what your concern is here.

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    #   (Note that the output needs to be independently verifiably correct.
+    #    Don't run an erroneous function, and then test against the erroneous
+    #    output you received when you ran your function for the first time.)
+    # Try this with important variations of parameters (reasonable, not
+    #    combinatorially exhaustive).
+    # Cover your code.
+
+    # ((Is graph.attributes(gWithThresh)$delta the correct way of retrieving
```
hyginn: **YOU** are writing these attributes so **YOU** know how to retrieve them from the results.

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    # delta values from graph attributes?)))
+    ### Check edgeThresh with different valid input values
+    ### For each check, length deltaVector should be Lmax - 1, numeric, and
+    ### between 0 and 1.
```
hyginn: Why between 0 and 1?

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    ### Check that the input graph is not changed
+    ### Test with N & Q = 1, 5, 10, Lmax 10 & 20
+    ### ((Test with N = 10, Q = 1 -> is there a limit to how many permutations
+    # there can be if only |E| edge swaps take place?))
+
+    gWithThresh <- edgeThresh(myEGG, Lmax = 20, N = 1, Q = 1)
+    deltaVector <- igraph::graph.attributes(gWithThresh)$delta
+
+    # Check that the input graph is not changed
```
hyginn: Can you explain to me by what computational process the input graph **could*** have been changed?

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    # ((what is the fastest way to check this?))
+    expect_true(isomorphic(gWithThresh, myEGG))
+
```
hyginn: Hang on: are you claiming that your edge permutations produce isomorphic graphs?

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    expect_equal(length(deltaVector), 9)
+    expect_equal(mode(deltaVector), "numeric")
+    expect_true(all(deltaVector[] >  0))
+    expect_true(all(deltaVector[] <= 1))
+
+})
+
+
+test_that("a corrupt input to edgeThresh does not lead to corrupted output", {
```
hyginn: Is that really necessary when working with an object that was prodcued by an internal function

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    # Try: - spurious characters in numeric columns ("N/A", ...).
+    #      - extra tabs at line-end
+    #      - comments before headers
+    #      - truncated file (incomplete last line)
+    # Again: make absolutely sure be you never have an erroneous
+    #        one-time execution of a loop
+
+    # Test with invalid EGG graphs
+    # ((How can one create an invalid EGG graph?))
```
hyginn: What assumptions are you making about the graph? An invalid graph is one that violates your assumptions.

tests/testthat/testEdgeThresh.R
```diff
@@ -0,0 +1,251 @@
...
+    invalidEGG1 <- igraph::rewire(myEGG, keeping_degseq(loops = TRUE))
```
hyginn: What are you doing here? What are you testing for?

tests/testthat/testEdgeThresh.R
FarzanT: Checking that invalid Lmax values are handled properly. If the size of the largest component is smaller than the given Lmax (for strongly connected components), I decided to return an error, since those values cannot result in distinct delta values. Isn't this appropriate?

tests/testthat/testEdgeThresh.R
FarzanT: That .permuteGraph() returns a permuted graph that maintains the number of vertices and edges, has no loops, multiple edges nor disconnected components.

FarzanT: Sorry I cannot see which line of code you are referring to exactly, but if you meant permuteGraph() parameter errors, this will test that invalid input arguments are handled properly.

tests/testthat/testEdgeThresh.R
FarzanT: But I thought that all EGG graphs should be simple.

FarzanT: Oh so should I also have a expect_false() condition where a non simple graph results in an error?

tests/testthat/testEdgeThresh.R
FarzanT: Now I need an explanation as to why I shouldn't have these tests :-))

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn Ok, I thought we have a standard for attaching attributes to graph objects in this package, wanted to make sure I was following that.

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn Sorry I meant the value for each delta should be between 0 and 1. Isn't this true?

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn Sorry I wasn't considering that only a __copy__ is passed to the function, I will remove this test.

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn No, I was looking for a fast way to ensure that the __original__ EGG that is passed to edgeThresh is not changed, but rather it should only have delta attributes attached. I thought that checking for isomorphism is the simplest way to do this? Is there a better way?

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn Shouldn't the number of __all__ edges stay the same?

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn Again, lines 170 to 214 are shown to me, so I don't know what is meant by _that_.
Shouldn't both .permuteGraph() as well as edgeThresh's return values be checked for correctness?

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn One that is not simple, does not satisfy rete.EGGprototype (still don't know what that will cover), is not directed (!) etc.
The etc depends on what rete.EGGprotype handles, e.g. valid gene names and valid attributes.

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn I am creating an invalid EGG, especially by creating a graph with the same nodes, but one that can have loops, making the EGG invalid.

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn Upon a second review, I am not checking if the input is mutated, but ensuring that what is return is the same graph but with new delta attributes attached, I will keep this test.

tests/testthat/testEdgeThresh.R
hyginn: Why are you not answering the question? First you are questioning whether the question is valid, thereby contradicting the reason of your test. Then you are reading an implication of a new requirement into my question. Just answer the simple question: what happens if the graph is not simple?

tests/testthat/testEdgeThresh.R
hyginn: You will **not** check whether the graph is **isomorphic**. That is (probably) an NP-complete problem. You will simply check whether the graphs are identical. That is done in *O*(|E|) by comparing the edges.

tests/testthat/testEdgeThresh.R
hyginn: Do you know what is meant when I ask about assumptions of an algorithm?

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn ~~A blackhole forms~~ Nothing really, since .permuteGraph() still has to ensure that __new__ edges formed after a swap do not result in self- or multi-edges. Also increasing delta wouldn't have an effect on self-edges, and multiple edges would make two vertices more resistant to edge removal if the edges have differing weights.

tests/testthat/testEdgeThresh.R
FarzanT: Certainly.

tests/testthat/testEdgeThresh.R
FarzanT: @hyginn Sorry, I listed how a graph would violate my assumption.
My assumptions are exactly the opposite of what I stated: 
-The graph __must__ be simple (that is what I thought an EGG should be). 
-The graph must satisfy rete.EGGprototype
-The graph is directed (not sure)
-It is __weighted__
-(Does not have delta attributes already? Not very important)
-Has max component order >= Lmax (as described in another comment)

PR #36, opened 2017-03-23
Merge pull request #1 from hyginn/master

bagherig:
```
> test_that("isGeneSymbol identifies invalid gene symbols", {
+     # Input contains invalid (non-HGNC) gene symbols.
+     input <- c("", "!@#", "AR3bsdt!", "123")
+     expected_output <- c(FALSE, FALSE, FALSE, FALSE)
+     expect_equal(isGeneSymbol(input), expected_output)
+ 
+ })
> test_that("isGeneSymbol identifies correct gene symbols, and input is case insensitive", {
+     # Input contains correct HGNC gene symbols.
+     input <- c("A1BG", "a1bg", "A1bG", "a1Bg")
+     expected_output <- c(TRUE, TRUE, TRUE, TRUE)
+     expect_equal(isGeneSymbol(input), expected_output)
+ })
> test_that("a corrupt input does not lead to corrupted output", {
+     # Input is corrupted (NULL, NA, vector of length zero, or no input at all).
+     expect_error(isGeneSymbol())
+     expect_error(isGeneSymbol(c()))
+     expect_error(isGeneSymbol(NULL))
+     expect_error(isGeneSymbol(NA))
+ })
```


R/checkGeneSymbols.R
```diff
@@ -0,0 +1,48 @@
...
+isGeneSymbol <- function(input) {
+    # Check the validity of the input.
+    if (is.null(input) || length(input) == 0 || typeof(input) != "character" ||
+        mode(input) != "character" || class(input) != "character") {
+            stop("Invalid input!")
+    }
+
+    output = c()
+
+    # Check each component of the input vector.
```
hyginn: Now you're giving up all the advantages of fast matching when you are iterating over the input, and worse, dynamiclly growing the output. This will be **really** slow. Write this as:
```
return(!is.na(fmatch(toupper(symbol), geneNames$Gene_Symbol)))
```
... and just for fun, try a microbenchmark on both versions with, say, 1,000 symbols and let me know hao much faster the second version is.

inst/scripts/generateHGNCHash.R
```diff
@@ -0,0 +1,59 @@
...
+#'
+#' @seealso \code{\link{isGeneSymbol}} is used to check if a gene symbol is a
+#' valid HGNC gene symbol that is contained in inst/extdata/HGNCSymbols.rds.
+
+library(fastmatch)
+
+# Open the connection/file.
+## Change fURL to be the URL of the file containing HGNC gene symbols.
+## The file is assumed to contain the symbol, status and locus group of the gene.
+fURL <- "http://tinyurl.com/kk3r9ck"
```
hyginn: No, that won't pass - you need to use the actile URL, otherwise, how would someone know that you are not taking them to an illicit Website...

inst/scripts/generateHGNCHash.R
```diff
@@ -0,0 +1,59 @@
...
+
+# Find the index of the columns that contain the gene symbol, status, and locus
+# group of the genes.
+symbolCol <- which(headerVector == "Approved Symbol")
+statusCol <- which(headerVector == "Status")
+locusCol <- which(headerVector == "Locus Group")
+
+# vectors containing the (important) statuses and locus groups of interest.
+impStatuses <- c("Approved")
+impLocusGroups <- c("protein-coding gene", "other", "phenotype")
```
hyginn: The selection we discussed per eMail included much more detail in LocusTypes, and it did not use LocusGroups. Why this change?

inst/scripts/generateHGNCHash.R
```diff
@@ -0,0 +1,59 @@
...
+
+# Read line by line.
+geneSymbols <- c()
+while (length(entry <- readLines(HGNCFile, n = 1)) > 0) {
```
hyginn: While this will work, (and I'm not requiring you to change this) the "proper" way here would be to use read.delim(), or better readr::read.delim(), select rows to include or exclude, and subset your symbols with a single statement.  No iteration.

tests/testthat/testCheckGeneSymbols.R
```diff
@@ -0,0 +1,67 @@
+# testDummy.R
+#
+# Dummy tests as a template for unit testing in test-driven development
```
hyginn: Delete comments from file templates unless you need them to explain the contents of **THIS** file.

tests/testthat/testCheckGeneSymbols.R
```diff
@@ -0,0 +1,67 @@
...
+#    test_that("a sane input gives an expected output", { ... })
+#    test_that("a corrupt input does not lead to corrupted output", { ...})
+
+
+context("checkGeneSymbols functions")
+
+
+# ==== BEGIN SETUP AND PREPARE =================================================
+OLOG <- as.character(getOption("rete.logfile"))   # save original logfile name
+logFileName(fPath = tempdir(), setOption = TRUE)  # make tempdir() the log dir
```
hyginn: Unnecessary, since you are not creating a log-file.

tests/testthat/testCheckGeneSymbols.R
```diff
@@ -0,0 +1,67 @@
...
+NL <- .PlatformLineBreak()
+# ==== END SETUP AND PREPARE ===================================================
+
+test_that("isGeneSymbol identifies invalid input", {
+    # Input contains invalid input.
+    input <- c(TRUE, 3.14, data.frame(key=c(1,2)), c(1,2,3))
+    expect_error(isGeneSymbol())
+    expect_error(isGeneSymbol(c()))
+    expect_error(isGeneSymbol(NULL))
+    expect_error(isGeneSymbol(input))
```
hyginn: Careful - there might be a symbol called input in the Global Environment. That's quite possible. Better to use something like noSuchObject . Thi also makes it more readable what you are testing for.

tests/testthat/testCheckGeneSymbols.R
```diff
@@ -0,0 +1,67 @@
...
+    expect_error(isGeneSymbol(TRUE))
+    expect_error(isGeneSymbol(3.14))
+    expect_error(isGeneSymbol(data.frame(key=c(1,2))))
+    expect_error(isGeneSymbol(c(1,2,3)))
+    expect_error(isGeneSymbol(NA))
```
hyginn: On the fence about this one - should this not just return FALSE? I think it should.

tests/testthat/testCheckGeneSymbols.R
```diff
@@ -0,0 +1,67 @@
...
+})
+
+
+test_that("isGeneSymbol identifies correct gene symbols,
+          and input is case insensitive", {
+    # Input contains correct HGNC gene symbols.
+    input <- c("A1BG", "a1bg", "A1bG", "a1Bg")
+    expected_output <- c(TRUE, TRUE, TRUE, TRUE)
+    expect_equal(isGeneSymbol(input), expected_output)
```
hyginn: Simply:
```
expect_equal(isGeneSymbol(c("A1BG", "a1bg", "A1bG", "a1Bg")),
                          c(TRUE, TRUE, TRUE, TRUE))
```

tests/testthat/testCheckGeneSymbols.R
```diff
@@ -0,0 +1,67 @@
...
+})
+
+
+test_that("isGeneSymbol identifies incorrect gene symbols
+          that are not present in the table", {
+    # Input includes symbols not present in the table.
+    input <- c("1234", "!@#$@", NA)
+    expected_output <- c(FALSE, FALSE, FALSE)
+    expect_equal(isGeneSymbol(input), expected_output)
```
hyginn: Simply:
```
expect_equal(isGeneSymbol(c("1234", "!@#$@", NA, "A1BG")),
                          c(FALSE, FALSE, FALSE, TRUE))
```

tests/testthat/testCheckGeneSymbols.R
bagherig: done

tests/testthat/testCheckGeneSymbols.R
bagherig: done

tests/testthat/testCheckGeneSymbols.R
bagherig: done

tests/testthat/testCheckGeneSymbols.R
bagherig: done

tests/testthat/testCheckGeneSymbols.R
bagherig: done

tests/testthat/testCheckGeneSymbols.R
bagherig: If you mean isGeneSymbol(NA), the problem with that is that:
```
> typeof(NA)
[1] "logical"
> class(NA)
[1] "logical"
> mode(NA)
[1] "logical"
```
So should I explicitly code so that if the input it NA, then return `FALSE`?

tests/testthat/testCheckGeneSymbols.R
bagherig: Okay I just thought my way would be more readable.
done

tests/testthat/testCheckGeneSymbols.R
bagherig: done

tests/testthat/testCheckGeneSymbols.R
bagherig: done

inst/scripts/generateHGNCHash.R
bagherig: done, I think (If I understood what you mean by **actile** URL correctly).
Could you clarify what you mean by an **actile** URL please?

inst/scripts/generateHGNCHash.R
bagherig: Well we never discussed which Locus Types to include. And in the email I said I think I should include locus groups "other" and "phenotype".

Here is the email, for reference:

> > Okay I think I get it. Also, here is what I got after calling unique on the different columns of the data obtained from HGNC.
> 
> > unique(status) 
> [1] "Approved"         "Symbol Withdrawn" "Entry Withdrawn" 
> > unique(locusType)
> [1] "gene with protein product"  "RNA, long non-coding"       "withdrawn"                  "pseudogene"          
> [5] "virus integration site"     "readthrough"                "phenotype only"             "unknown"                   
> [9] "region"                     "endogenous retrovirus"      "fragile site"               "immunoglobulin gene"       
> [13] "immunoglobulin pseudogene"  "transposable element"       "RNA, micro"                 "RNA, ribosomal"            
> [17] "RNA, transfer"              "complex locus constituent"  "protocadherin"              "RNA, cluster"              
> [21] "RNA, misc"                  "RNA, small nuclear"         "RNA, small cytoplasmic"     "RNA, small nucleolar"      
> [25] "RNA, Y"                     "T-cell receptor gene"       "T-cell receptor pseudogene" "RNA, vault"                
> > unique(locusGroup)
> [1] "protein-coding gene" "non-coding RNA"      "withdrawn"           "pseudogene"          "other"               
> [6] "phenotype"          
>
> Which genes would you say I should include in the data frame? I think the genes with Locus Group "other" and "phenotype" might also be important? What do you think?

bagherig: So which Locus Types should I include in the data frame? I think the pseudogenes should NOT be included. But the following should be included:
> "gene with protein product"
> "virus integration site"
> "region"
> "fragile site"
> "endogenous retrovirus"
> complex locus constituent 
> immunoglobulin gene
> phenotype only
> protocadherin 
> readthrough 
> T cell receptor gene 
> transposable element

Also, perhaps the RNA ones are also important? What do you think? [Here](http://www.genenames.org/help/download) is a list of definitions of the terms, in case you want to look at them.

Thanks

R/checkGeneSymbols.R
bagherig: wow you are right I did not know you can fmatch a vector against a vector...

Here are the microbenchmark results:
```
> microbenchmark(isGeneSymbol(a), isGeneSymbol2(a))
Unit: microseconds
             expr      min         lq       mean    median        uq       max neval cld
  isGeneSymbol(a) 9792.464 10463.5370 11418.6488 10828.522 11501.450 16326.030   100   b
 isGeneSymbol2(a)  145.160   154.8995   194.1434   185.972   224.001   370.088   100  a 
```
isGeneSymbol is my version and isGeneSymbol2 is your version. Your version is much much faster!
Also, the input "a" included 1002 components.

tests/testthat/testCheckGeneSymbols.R
hyginn: Hm. You an also do ```x <- as.character(NA)``` and then ```mode(x)``` is "character". Now that makes me worried that we would do so much testing here, that the fast function would be made slow again. By comparing the behaviour of match() with is.integer, I think we should do the following: if the input is NULL or length() == 0, we return FALSE:
```
if (length(input) == 0 || is.null(input) { return(FALSE) }
```
because neither of the two is a gene symbol. Otherwise we cast the input to character (which is fast, if it already character in the first place), convert to upper, and ```fmatch()``` it. Almost everything will be flattened into a vector of elements and processed. A single NA returns FALSE, and all nonsense that I could think of and tried does too.  That way we can virtually guarantee that the function returns a logical - and for example - that makes it safe to use in an ```if()``` statement. Thus the actual function would be:
```
return(!is.na(fmatch(toupper(as.character(input), geneNames$Gene_Symbol)))
```
... and if the user chooses to feed the function a bunch of numbers, that's none of our business - we'll just return a bunch of FALSE - which is correct because a number is not a gene symbol.

inst/scripts/generateHGNCHash.R
hyginn: I did send you a list of locusTypes - but NVM, I can just make the change. No big deal.

hyginn:
I think it's all done. Smooth sailing overall. I'll merge it in, then apply the few polishing steps that I still thought of. Excellent.

hyginn:
Merged.

PR #71, opened 2017-04-01
importPDB.R pseudocode

bagherig:
I think the pseudocode is done now, except for a couple of things that I was unsure of which I have mentioned in the comments. Thanks.


inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
+library("bio3d")
```
hyginn: Don't put ```library()``` calls into your code

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+library("Biostrings")
+install.packages("seqnir")
```
hyginn: Typo

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+
+readRDS(MT.RDS)
+
+
+for (genes in gene_list){
+    results <- list()
+    # Translate the gene to its aa sequence
```
hyginn: Check first whether there is a structure map at all.

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+        sidechainXYZ <- c()
+        calphaXYZ <- c()
+
+        # load the PDB file.
+        subject <- read.pdb(PDB-ID)
+        # get the aa sequence of the subject
+        subjectAA <- AAString(paste(PDB$seqres, collapse=""))
+        #Perform pairwisealignment..
+        ali <- pairwiseAlignment(subjectAA, targetAA)
+        # ^^^ Q: what should the gapExtension and gapOpening penalties be?
```
hyginn: Use default values for matrix / opening / extension. In this scenario you expect 100% identity...

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+        for (i in 1:nmatch(ali)){   #For all matched residues:
+            i <- index of residue in targetAA
+            j <- index of residue in subjectAA
```
hyginn: I don't know. That's for you to figure out.

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+        rowNames <- c()
+        resIDs <- c()
+        aaType <- c()
+        sidechainXYZ <- c()
+        calphaXYZ <- c()
+
+        # load the PDB file.
+        subject <- read.pdb(PDB-ID)
+        # get the aa sequence of the subject
+        subjectAA <- AAString(paste(PDB$seqres, collapse=""))
```
hyginn: Note that you need to work with the right chain!

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+        #Perform pairwisealignment..
+        ali <- pairwiseAlignment(subjectAA, targetAA)
+        # ^^^ Q: what should the gapExtension and gapOpening penalties be?
+        for (i in 1:nmatch(ali)){   #For all matched residues:
+            i <- index of residue in targetAA
+            j <- index of residue in subjectAA
+            #^^^ Q: How do I get the indices of residues aligned in target and subject?
+            rowNames <- append(rowNames, i)     # rowname will be the index of the residue in target gene
```
hyginn: That looks fishy - you are dynamically growing this vector ... can't you construct it in one expression before you enter the loop?

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+            resIDs <- append(unique(subject$atom[subject$atom$resno==j,]$resid)) #resids are the 3 letter code of the aa
```
hyginn: Why do you need that?

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+            aaTypes <- append(aaTypes, type of targetAA[i])
```
hyginn: Why do you need that?

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+            # ^^^ Q: what is aaType? (hydrophobic vs. hydrophilic? or acidic vs. basic? or something else?)
+
+            # specify which atoms in subject$atom$elety are for alpha carbon, COOH and NH2.
+            # The rest of the atoms must be in the side chain
+            nonSideChain_elety <- c("N", "C", "O", "H", "CA")
```
hyginn: Not sure - is it really called "elety"? Be careful: you are looking for atom names, not chemical element type.

Also: skip all hydrogen atoms.

Also: sometimes atoms and residues are listed multiple times in alternate locations. Make sure your code handles that correctly. I don't know how Bio3D flags that.

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+            thisResData <- subject$atom[subject$atom$resno== j,]
```
hyginn: please consistently put spaces on both sides of an operator. ```resno == j```

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+
+            #If there is data for this residue:
+            if (nrow(thisResData)!=0){
+                #x, y, z coordinates of the side chain atoms
+                sideChainXValues <- thisResData[!(thisResData$elety %in% nonSideChain_elety), "x"]
```
hyginn: I'm pretty sure there is a more concise way to write this directly from subsetting the data vector, rather than searching for matches with an %in% expression.

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+                sideChainYValues <- thisResData[!(thisResData$elety %in% nonSideChain_elety), "y"]
+                sideChainZValues <- thisResData[!(thisResData$elety %in% nonSideChain_elety), "z"]
+
+                # calculate the average of the x, y, z coordinates
+                avgX <- sum(sideChainXValues)/length(sideChainXValues)
```
hyginn: ```mean()``` ???

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+                avgY <- sum(sideChainYValues)/length(sideChainYValues)
+                avgZ <- sum(sideChainZValues)/length(sideChainZValues)
+                sidechainXYZ <- append(sidechainXYZ, c(avgX, avgY, avgZ))
```
hyginn: Again ```append()```. Grrr.

Don't. Dynamically. Grow. Vectors.

(Unless in the rare cases when you have to. This is not one of them.)

Who teaches you this anyway? I'm pretty sure **none** of my code ever uses append.

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,76 @@
...
+
+                # x, y, z coordinates of the alpha carbon
+                calphaXYZ <- append(calphaXYZ, c(thisResData[thisResData$elety=="CA", "x"),
```
hyginn: Again, subset, then assign the whole row at once ...
Consider:
```
(myTest <- matrix(numeric(12), ncol = 3))
(myCA <- matrix(runif(12), ncol = 3))

myTest[3, ] <- myCA[1,]
myTest
```

bagherig:
It appears I forgot to push my code before I submitted a pull request.. because this isn't my latest version of the pseudocode.. I'll fix what you have mentioned and submit again

inst/scripts/importPDB.R
bagherig: Well the description of GeneData$3D on the task page says "Columns are the the **structure residue ID**, **amino acid type**, and the x, y, and z coordinates of the Cα atom and of the side-chain centroid.

inst/scripts/importPDB.R
bagherig: see above ^

inst/scripts/importPDB.R
bagherig: I tried subsetting the data frame using the value of two columns.. but it seems it is not possible.. I could only subset using the value of one column only.. I will try again

inst/scripts/importPDB.R
bagherig: done

inst/scripts/importPDB.R
bagherig: I just couldn't figure a way to create an empty data frame with column names but no values and then adding to the data frame.. because once you add the first row of values using rbind, R changes the column names to x1, x2, etc...

inst/scripts/importPDB.R
bagherig: done

inst/scripts/importPDB.R
bagherig: done

inst/scripts/importPDB.R
bagherig: done

inst/scripts/importPDB.R
bagherig: done, i think..

inst/scripts/importPDB.R
bagherig: But wouldn't I need the index of the residue ("i") from the for-loop?

inst/scripts/importPDB.R
bagherig: done

inst/scripts/importPDB.R
bagherig: done

inst/scripts/importPDB.R
bagherig: done

inst/scripts/importPDB.R
bagherig: done, i think..

PR #74, opened 2017-04-01
SEL3D milestone 3 "pseudocode"

LindsayLiang:
....I  may have gotten carried away and just wrote actual code...I hope that's ok.

### DIFF
--- a/inst/scripts/SEL3D.R
+++ b/inst/scripts/SEL3D.R
@@ -1,0 +1,108 @@
+#Pseudocode may look like actual code...please ignore...
+get_pdb_ids <- function(target_seq, start_pos = 1 , end_pos = nchar(target_seq), convert = TRUE, silent = TRUE){
+  
+  #get packages
+  # install.packages("bio3d")
+  # install.packages("RCurl")
+  # install.packages("XML")
+  # install.packages("seqinr")
+  # library(bio3d)
+  # library(RCurl)
+  # library(XML)
+  # library(seqinr)
+  
+  
+  
+  #convert sequence into AA 
+  if(convert){
+    #translate needs a vector of single chars, not a string
+    AA_seq <- paste(translate(s2c(target_seq)), collapse='')
+    end_pos <- nchar(AA_seq)
+  }else{ AA_seq <- target_seq }
+ 
+  blast_output <- blast.pdb(AA_seq)
+  
+  blast_output <- blast_output$hit.tbl
+  blast_output <- blast_output[blast_output$identity > 40, ]
+  
+  if(nrow(blast_output) == 0){
+    return(NULL)
+  }
+  
+  blast_output$tax <- NA 
+  blast_output$info <- NA
+  
+  final_output <- data.frame()
+  final_output_range <- c()
+  final_output_range_counter <- 1
+  
+  for(entry in 1:nrow(blast_output)){
+    rcsb_id <- gsub("_", ".", blast_output$pdb.id[entry], ignore.case = TRUE)
+    lookup <- xmlParse(getURL(paste("http://www.rcsb.org/pdb/rest/describeMol?structureId=", rcsb_id, sep = "")))
+    
+    #get taxonomy 
+    
+    tax <- getNodeSet(lookup, '//Taxonomy/@name')[[1]][[1]]
+    info <- getNodeSet(lookup, '//polymerDescription/@description')[[1]][[1]]
+    
+    blast_output$tax[entry] <- ifelse(length(tax) > 0, tax, NA)
+    blast_output$info[entry] <- ifelse(length(info) > 0, yes = info, no = NA)
+    
+    if(!intersecting_positions(final_output_range, blast_output$q.start[entry], blast_output$q.end[entry])){
+      if((!is.na(blast_output$tax[entry]) && blast_output$tax[entry] == "Homo sapiens") || blast_output$identity[entry] == 100){
+        final_output <- rbind(final_output, blast_output[entry,])
+        final_output_range[[final_output_range_counter]] <- c(blast_output$q.start[entry]:blast_output$q.end[entry])
+        final_output_range_counter <- final_output_range_counter + 1
+      }
+    }
+  }
+  
+  if(nrow(final_output) == 0){ return(NULL) }
+  
+  final_output$q.start <- final_output$q.start + start_pos - 1
+  final_output$q.end <- final_output$q.end + start_pos - 1
+  
+  for(x in 1:length(final_output_range)){
+    final_output_range[[x]] <- final_output_range[[x]] + start_pos - 1
+  }
+  
+  min_side <- min(final_output_range[[1]])
+  max_side <- max(final_output_range[[length(final_output_range)]])
+  
+  if(end_pos - max_side > 50){
+    message(sprintf("running substr from pos %d to %d", max_side, end_pos))
+    max_frame <- get_pdb_ids(substr(AA_seq, max_side - start_pos, end_pos - start_pos), 
+                             start_pos = max_side, end_pos = end_pos, convert = FALSE, silent = silent)
+    final_output <- rbind(final_output, max_frame)
+  }
+  
+  if(min_side > 50){
+    message(sprintf("running substr from pos %d to %d", start_pos, min_side))
+    min_frame <- get_pdb_ids(substr(AA_seq, 1, min_side - start_pos), 
+                             start_pos = start_pos, end_pos = min_side, convert = FALSE, silent = silent)
+    final_output <- rbind(final_output, min_frame)
+  }
+  
+  for(y in 1:(length(final_output_range) - 1)){
+    left_side <- max(final_output_range[[y]])
+    right_side <- min(final_output_range[[y+1]])
+    
+    if(right_side - left_side > 50){
+      message(sprintf("running substr from pos %d to %d", left_side, right_side))
+      inter_frame <- get_pdb_ids(substr(AA_seq, left_side - start_pos, right_side - start_pos), 
+                               start_pos = left_side, end_pos = right_side, convert = FALSE, silent = silent)
+      final_output <- rbind(final_output, inter_frame)
+    }
+  }
+  
+  return(final_output)
+}
+
+intersecting_positions <- function(r, st, ed){
+  for(x in 1:length(r)){
+    if(length(intersect(r[[x]], c(st:ed))) > 0){
+      return(TRUE)
+    }
+  }
+  return(FALSE)
+}


hyginn:
Just put in an Issue when you want it reviewed...
B.

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
+#Pseudocode may look like actual code...please ignore...
+get_pdb_ids <- function(target_seq, start_pos = 1 , end_pos = nchar(target_seq), convert = TRUE, silent = TRUE){
+  
+  #get packages
+  # install.packages("bio3d")
+  # install.packages("RCurl")
+  # install.packages("XML")
+  # install.packages("seqinr")
+  # library(bio3d)
+  # library(RCurl)
```
hyginn: What about httr?

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+  if(convert){
+    #translate needs a vector of single chars, not a string
+    AA_seq <- paste(translate(s2c(target_seq)), collapse='')
+    end_pos <- nchar(AA_seq)
+  }else{ AA_seq <- target_seq }
+ 
+  blast_output <- blast.pdb(AA_seq)
+  
+  blast_output <- blast_output$hit.tbl
+  blast_output <- blast_output[blast_output$identity > 40, ]
```
hyginn: Don't you need to align, then check **where** you actually have 100% ID? You might have 90% overall, but including malaligned fringes...

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+  
+  blast_output$tax <- NA 
+  blast_output$info <- NA
+  
+  final_output <- data.frame()
+  final_output_range <- c()
+  final_output_range_counter <- 1
+  
+  for(entry in 1:nrow(blast_output)){
+    rcsb_id <- gsub("_", ".", blast_output$pdb.id[entry], ignore.case = TRUE)
```
hyginn: See this is precisely where pseudocode would have been preferred: I happen to know what you are doing here, but for everyone else this is completely unnecessarily opaque. What's keeping you from writing?
```
# Reformat PDB ID as output by BLAST to separate ID from chain with a "."
```
This has nothing to do with getting "carried away" with writing code - and everything to do with not being explicit in this conversation. 

(Incidentally: why is this even necessary?)

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+    lookup <- xmlParse(getURL(paste("http://www.rcsb.org/pdb/rest/describeMol?structureId=", rcsb_id, sep = "")))
```
hyginn: Not a good choice of variable name. Has no semantics associated, and all lowercase sounds like a verb.

inst/scripts/SEL3D.R
LindsayLiang: I'm a bit confused, blast.pdb returns a table of aligned segments, not an overall alignment, so if you're worried about say if we had 90% alignment and the middle of the sequence contained poor sequence identity, I don't think this would happen.

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+    
+    #get taxonomy 
+    
+    tax <- getNodeSet(lookup, '//Taxonomy/@name')[[1]][[1]]
+    info <- getNodeSet(lookup, '//polymerDescription/@description')[[1]][[1]]
+    
+    blast_output$tax[entry] <- ifelse(length(tax) > 0, tax, NA)
```
hyginn: I discourage use of ```ifelse()```. Not good writing style, and not easy to maintain when the moment comes that you need to do more. Generally this makes code harder to read.

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+}
+
+intersecting_positions <- function(r, st, ed){
+  for(x in 1:length(r)){
+    if(length(intersect(r[[x]], c(st:ed))) > 0){
+      return(TRUE)
+    }
+  }
+  return(FALSE)
+}
```
hyginn: No newline at end of file. Add ```# [END]``` and make it a habit to reader other submission comments because I've discussed this elsewhere. If you're not part of the conversation in a collaborative project, it costs me extra time because I have to flag things over and over again, and it costs you extra time because many things could easily have been avoided.

inst/scripts/SEL3D.R
LindsayLiang: Necessary because we're looking up taxonomy and the description of the pdbID from a different server which uses slightly different formatting.

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+                               start_pos = left_side, end_pos = right_side, convert = FALSE, silent = silent)
+      final_output <- rbind(final_output, inter_frame)
+    }
+  }
+  
+  return(final_output)
+}
+
+intersecting_positions <- function(r, st, ed){
+  for(x in 1:length(r)){
```
hyginn: This looks weird. Can you explain what you are doing here, especially what data structure this works on?

inst/scripts/SEL3D.R
hyginn: How is this going to work with ModBase entries?

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+    #get taxonomy 
+    
+    tax <- getNodeSet(lookup, '//Taxonomy/@name')[[1]][[1]]
+    info <- getNodeSet(lookup, '//polymerDescription/@description')[[1]][[1]]
+    
+    blast_output$tax[entry] <- ifelse(length(tax) > 0, tax, NA)
+    blast_output$info[entry] <- ifelse(length(info) > 0, yes = info, no = NA)
+    
+    if(!intersecting_positions(final_output_range, blast_output$q.start[entry], blast_output$q.end[entry])){
+      if((!is.na(blast_output$tax[entry]) && blast_output$tax[entry] == "Homo sapiens") || blast_output$identity[entry] == 100){
```
hyginn: * Why is the taxonomy not the very first thing you test?
* Why are you searching for ID > 40 but requiring ID == 100 here?

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
+#Pseudocode may look like actual code...please ignore...
```
hyginn: I would prefer you to use camelCase, not pothole_case.

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+    tax <- getNodeSet(lookup, '//Taxonomy/@name')[[1]][[1]]
+    info <- getNodeSet(lookup, '//polymerDescription/@description')[[1]][[1]]
+    
+    blast_output$tax[entry] <- ifelse(length(tax) > 0, tax, NA)
+    blast_output$info[entry] <- ifelse(length(info) > 0, yes = info, no = NA)
+    
+    if(!intersecting_positions(final_output_range, blast_output$q.start[entry], blast_output$q.end[entry])){
+      if((!is.na(blast_output$tax[entry]) && blast_output$tax[entry] == "Homo sapiens") || blast_output$identity[entry] == 100){
+        final_output <- rbind(final_output, blast_output[entry,])
+        final_output_range[[final_output_range_counter]] <- c(blast_output$q.start[entry]:blast_output$q.end[entry])
```
hyginn: Poor choice of variable name: a "counter" sounds like a single number. You have a list of results. That's really confusing.

inst/scripts/SEL3D.R
LindsayLiang: We need to store ranges of the best alignments (as start (st) and end (ed) positions) so we can add that match to the final output without adding segments which overlap in the final format.

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+        final_output_range_counter <- final_output_range_counter + 1
+      }
+    }
+  }
+  
+  if(nrow(final_output) == 0){ return(NULL) }
```
hyginn: That's it? No comment? No record of how many failed to have matches? No record of whether the match failed in PDB and/or Modbase?

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+  
+  final_output$q.start <- final_output$q.start + start_pos - 1
```
hyginn: Why?

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,108 @@
...
+    right_side <- min(final_output_range[[y+1]])
+    
+    if(right_side - left_side > 50){
+      message(sprintf("running substr from pos %d to %d", left_side, right_side))
+      inter_frame <- get_pdb_ids(substr(AA_seq, left_side - start_pos, right_side - start_pos), 
+                               start_pos = left_side, end_pos = right_side, convert = FALSE, silent = silent)
+      final_output <- rbind(final_output, inter_frame)
+    }
+  }
+
```
hyginn: Where is ```Post-process the matches for each gene to minimize overlap.```

inst/scripts/SEL3D.R
LindsayLiang: For the recursion, if we blast a sequence, that's lest say 100aa long,  and we get matches for 1-10 and 90-100, but not matches for 11-89, we have to recurse on that new substring.  However, if we were just to reblast the subsequence, the start and end positions would be reset to 1 and 79, (ie the length of the sub string), which aren't actually the start and end positions of the original sequence, which is what we want.

inst/scripts/SEL3D.R
LindsayLiang: First we drop any match with ID < 40 to save runtime, then we query each of those matches for their tax and description.

We don't want to drop all the non-homo sapien matches because of the edge case where a match could have 100% amino acid ID for a non homo sapien which we want to include in the table (we discussed this in the notes section of the wikipage).  

We then find the best match that is either Homo sapiens or 100% ID.

inst/scripts/SEL3D.R
LindsayLiang: There is no overlap, we eliminate that by re-blasting subsets of the input sequence that have no matches anywhere (see above).

inst/scripts/SEL3D.R
LindsayLiang: Modbase's online tool currently doesn't work (page not found), so I'm hesitant to download the 40GB of files they have in their ftp site...

inst/scripts/SEL3D.R
LindsayLiang: httr was replaced by XML and RCurl which allow me to use XPath (which I wanted to apply here since I learned it in CSC343...)

PR #76, opened 2017-04-01
importPDB.R pseudocode

bagherig:


inst/scripts/importPDB.R
```diff
@@ -0,0 +1,158 @@
+# importPDB.R
+# Non-exported function
+
```
RicardoRam, line 3: Comments look good.

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,158 @@
...
+        chainAAIndices <- atom.select(subject,
+                                      "calpha",
+                                      chain = chainID,
+                                      value = TRUE)
+        subjectAA <- AAString(paste(aa321(chainAAIndices$atom$resid), collapse = ""))
+
+        #=====================ALIGNMENT========================================
+
+        # retrieve the start and end indices for the target & subject sequences:
+        start_target <- the starting index of the alignment in targetAA from MT
```
RicardoRam, line 72: You might want to keep pseudo-code in comments so file is runnable though.

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,158 @@
...
+        end_subject <- the last index of the alignment in target AA from MT
+
+        # perform pairwise alignment with default parameters.
+        ali <- pairwiseAlignment(subjectAA[start_subject, end_subject],
+                                 targetAA[start_target, end_target])
+
+        #======================================================================
+
+        # create a data frame with nmatch(ali) rows.
+        DF <- data.frame("ResID" = c(1:nmatch(ali)),
```
hyginn, line 84: What package is ```nmatch()``` from? And why are you calling it five times on the same data? Is it expensive to calculate? And why are you pre-loading these vectors with sequential integers rather than just using ```numeric(n)```?

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,158 @@
...
+            i <- index of residue in targetAA
+            j <- index of residue in subjectAA
+
+            # rowname will be the index of the residue in target gene:
+            rownames(DF)[n] <- i
+
+            #=================RESIDUE ID & AA TYPE=============================
+
+            # resids are the 3 letter code of the aa:
+            resID <- unique(subject$atom[subject$atom$resno==j,]$resid)
```
hyginn, line 100: Why ```unique()```? (Comment required!)

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,158 @@
...
+            # type of AA..
+            aaType <- type of targetAA[i]
+            # ^^^ Q: what does it mean by amino acid type?
+            #        (hydrophobic vs. hydrophilic? or acidic vs. basic? or
+            #        something else?) Still need to figure out why we need
+            #        these.. They are mentioned in the design task page of
+            #        the GeneData object.
```
hyginn, line 107: I don't see "amino acid type" on the geneData task page ... what are you referring to?

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,158 @@
+# importPDB.R
+# Non-exported function
+
+#=============================STOP & SAVE======================================
+# Provide a mechanism to stop:
```
hyginn, line 5: Interesting ...

inst/scripts/importPDB.R
bagherig: [GeneData resource](http://steipe.biochemistry.utoronto.ca/abc/students/index.php/User:Sabrina.ge/Tasks/SNV_Genes_data_struct)
In the table where `GeneData` is described, row `$3D`, column `semantics`: 
> list indices are PDB-ID/chain. For each PDB-ID/chain there is a data frame in which rownames are the gene's sequence positions for which 3D coordinates have been mapped to this structure. Columns are the the **structure residue ID, amino acid type**, and the x, y, and z coordinates of the Cα atom and of the side-chain centroid.

inst/scripts/importPDB.R
bagherig: because I am subsetting all atoms that have `$resno==j`, and then I'm getting their `$resid`. This will give me multiple instances of the same amino acid. for example: 
`[1] "HIS" "HIS" "HIS" "HIS" "HIS" "HIS" "HIS" "HIS" "HIS" "HIS" "HIS" "HIS" "HIS" "HIS" "HIS"`
Calling unique will retrieve only one of them  (I could alternatively subset the first element using `[1]` ). However, I just realized that `$resno` resets for each chain, so I will need to make sure that I only find the `$resno` in the desired chain or else I will get multiple amino acids, one for each chain.
I will comment this on the code.

inst/scripts/importPDB.R
bagherig: Do you think this will work as a way of stopping the process and continuing it later? I think it will! It is hard to test some things though as I don't have actual data to work with.
Or do you rather not have a GUI for this?

inst/scripts/importPDB.R
bagherig: you are right, I should call it once and store the value in a variable and then use that variable five times. But I did this because I needed the number of rows of the data frame to be the same as the number of residues matched. This way, I can just change each row after each iteration to have the correct values.
Also, If I just created an empty data frame and add to it, R would automatically change the column names to x1, x2, etc... 
But I think I just figured that I can maybe change the column names at the end, when all the for loops are done computing. I will have to try this and see if it works.

bagherig: And `nmatch()` is from the `Biostrings` package. After doing `pairwiseAlignment()`, you can call `nmatch()` on the alignment object to get the number of residues that were aligned.

inst/scripts/importPDB.R
hyginn: I don't know yet. It's an interesting idea. Leave it in so we can play with it. We can always comment it out.

inst/scripts/importPDB.R
```diff
@@ -0,0 +1,158 @@
...
+                    })
+tkgrid(butStop)
+tkgrid(butSave)
+stoploop <- FALSE
+
+readRDS(MT.rds)
+
+#=============================(LOAD & CONTINUE)/START==========================
+# Check if there was a saved progress from before.
+if (file.exists("inst/tmp/genes.rds")){
```
hyginn, line 32: It's not a good idea to hardcode a file-path like that. Use a variable ```PROGRESSFILE``` that you define at the top of the script, and **tell the user** that you are doing this. Worst case scenario: the user forgot about the progress file for a different project and your code silently patches two incompatible lists together.

inst/scripts/importPDB.R
hyginn: Write this like so:
```
nMatch <- nmatch(ali)
DF <- data.frame(ResID = numeric(nMatch),  # Is this numeric? Or is there possibly an insert code attached?
                 AA = character(nMatch),
                 xyzSC = numeric(nMatch),
                 xyzCA = numeric(nMatch),
                 stringsAsFactors = FALSE))    # Do. Not. Forget. This.
```

inst/scripts/importPDB.R
hyginn: I see. But there ought to be a vector of residue IDs in the PDB object ...

Also be careful of the fact that there may be insert codes. For an example, see residues 31, 31A, 31B etc. in 2IMM. Unique residue "names" are not numbers.

inst/scripts/importPDB.R
hyginn: Ah. One of the twenty one-letter amino acid codes.

hyginn:
One general comment: use ```bio3d::function()```, biostrings::function()``` etc. - it makes the code much more maintainable; especially with packages from related domains its often not obvious which package a function comes from.

inst/scripts/importPDB.R
bagherig: done.
But would you think the other way that I mentioned might be better and possible faster? 

> But I think I just figured that I can maybe change the column names at the end, when all the for loops are done computing. I will have to try this and see if it works.

But I think it might look a bit weird/confusing in the code, since I will have to name the columns twice: once when I create the empty data frame, and once after the data frame has been populated.

inst/scripts/importPDB.R
bagherig: > One of the twenty one-letter amino acid codes.

Is that the resID or aaType?
then what would the other one be? because I already have resID as the 3-letter amino acid code.

inst/scripts/importPDB.R
bagherig: done. 
I changed my code to include the _insert code_

inst/scripts/importPDB.R
RicardoRam: Hmm. I think resID is correct. Here's a link on PDB of the definition.
http://www.rcsb.org/pdb/staticHelp.do?p=help/advancedsearch/residueID.html
I couldn't find out what aatype was either. I also don't see the point of storing both the 3 letter and 1 letter name for amino acids.

PR #82, opened 2017-04-02
SEL3D with function description, input validation, and comments for clarity.

LindsayLiang:
Added: Function description, input validation, more comments, required packages to DESCRIPTION file.

### DIFF
--- a/inst/scripts/SEL3D.R
+++ b/inst/scripts/SEL3D.R
@@ -145,10 +145,9 @@ getPdbIds <- function(targetSeq, startPos = 1 , endPos = 1, convert = TRUE, sile
   
     return(finalOutput)
   }
-  #We need to store ranges of the best alignments 
-  #(as start (st) and end (ed) positions) so we can
-  #add that match to the final output without adding
-  #segments which overlap in the final format.
+  #We need to store ranges of the best alignments (as start (st) and end (ed) positions)
+  #so we can add that match to the final output without adding segments which overlap
+  #in the final format.
   intersectingPositions <- function(r, st, ed){
     for(x in 1:length(r)){
       if(length(intersect(r[[x]], c(st:ed))) > 0){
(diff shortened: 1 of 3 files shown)


inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,160 @@
...
+      finalOutput <- rbind(finalOutput, interFrame)
+    }
+  }
+  
+    return(finalOutput)
+  }
+  #We need to store ranges of the best alignments 
+  #(as start (st) and end (ed) positions) so we can
+  #add that match to the final output without adding
+  #segments which overlap in the final format.
```
RicardoRam: Please follow the function description convention in the /R folder

inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,160 @@
+getPdbIds <- function(targetSeq, startPos = 1 , endPos = 1, convert = TRUE, silent = TRUE){
```
RicardoRam: Please follow the function description convention in the /R folder

tests/testthat/testSel3D.R
```diff
@@ -0,0 +1,55 @@
...
+#NL <- .PlatformLineBreak()
+# ==== END SETUP AND PREPARE ===================================================
+
+
+
+
+#test_that("parameter errors are correctly handled", {
+  # Try missing input sequence
+  # Try extraneous parameters
+  # Note: we're not to attempt to avoid all kinds of abuse from users
```
RicardoRam: In my opinion, if the parameters are just for recursion, you could:
1. Explicitly tell the user these parameters are preset and not meant to be changed in the preset argument descriptions. I think this option is "code smell" though.
2. Have getPdbIds set up a helper function that handles the recursive part. If the parameters aren't supposed to be changed, I think exposing them to the user should be avoided. 

What's your opinion?

LindsayLiang: Yes a wrapper would work, but it does seem a bit silly.  How would one 'explicitly' tell the user these things, other than just putting the information in the file header?

RicardoRam: The following link has information about proper documentation. 
http://r-pkgs.had.co.nz/man.html
In particular, 
"@param name description describes the function’s inputs or parameters. The description should provide a succinct summary of the type of the parameter (e.g., string, numeric vector) and, if not obvious from the name, what the parameter does."

RicardoRam: It might be easier to refer to the R code that has been merged into the rete package when filling out your function documentation. Don't "reinvent the wheel".

PR #84, opened 2017-04-02
importMapCoord.R

Sabrina-Ge:
In the order they appear in the file:
makeMap(): single gene model function
Validation workflow code + description for makeMap
mapCoord():  function factory prototype


inst/scripts/.importMapCoord.R
```diff
@@ -0,0 +1,169 @@
+# importMapCoord.R
+# non-exported
+
+#==========================makeMap and Validation===============================
+
+# Helper function - generates the closure for a single gene model
+makeMap <- function(ucscMod) {
+
+    # Check if the number of exons listed in start and end match
+    if(length(ucscMod$exonStart) != length(ucscMod$exonEnd)) {
```
hyginn: Since this situation should never occur, its better to actually put in a ```stop("Corrupt gene model.")```

inst/scripts/.importMapCoord.R
```diff
@@ -0,0 +1,169 @@
...
+    start <- ucscMod$start
+    end <- ucscMod$end
+    reverse <- FALSE
+    if (ucscMod$strand == "-") {
+        reverse <- TRUE
+    }
+
+    # Check if the exons and start stop points are validly ordered
+    if(is.unsorted(c(start, as.vector(t(exons)), end))) {
+        return(NULL)
```
hyginn: As above ... ```stop()``` with error message.

inst/scripts/.importMapCoord.R
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    # Return a function that uses the exon matrix, strand and start, stop positions
```
hyginn: How about pre-calculating a third column with cumulative exon-length sums  ```cumsum()``` so you don't need to do the sums in every function call. You could just then just use ```exons$cumsum[iEx] - (exons$end[iEx] - x)``` (probably need to add or subtract a 1 somewhere in there) to get the position. Also you could switch start and end for (-) strand, when you prepare ```exons```, then you don't have to special-case the (-)-strand.

Sabrina-Ge:
The exon sum method works and makes the code look nicer. It takes longer to generate the map function but the map function then runs faster for all coordinate inputs.

However, I could not find a way to generate the map function without passing in the strand since. Even if I mutated the exon table, I still needed to "flip" the x-coordinate when the strand is (-) to get an accurate position or change what to subtract and add.

hyginn:
> On Apr 2, 2017, at 7:23 PM, Sabrina-Ge <notifications@github.com> wrote:
> 
> The exon sum method works and makes the code look nicer.

That's great - that's a win on all counts then.

> It takes longer to generate the map function but the map function then runs faster for all coordinate inputs.
> 
> However, I could not find a way to generate the map function without passing in the strand since.

Ah, a puzzle - can I see the code?

B.

Sabrina-Ge:
```
makeMap <- function(ucscMod) {

    # Check if the number of exons listed in start and end match
    if(length(ucscMod$exonStart) != length(ucscMod$exonEnd)) {
        stop("Corrupt gene model")
    }

    # Prepare the environment (exon matrix, start, end, strand)
    exons <- cbind(ucscMod$exonStart, ucscMod$exonEnd)
    start <- ucscMod$start
    end <- ucscMod$end
    reverse <- FALSE
    if (ucscMod$strand == "-") {
        reverse <- TRUE
    }

    # Check if the exons and start stop points are validly ordered
    if(is.unsorted(c(start, as.vector(t(exons)), end))) {
        stop("Corrupt gene model")
    }

    # Generate cumulative sums
    cumuSums <- vector(mode="numeric", length=nrow(exons))
    for (i in 1:nrow(exons)) {
        cumuSums[i] <- sum((exons[ , 2] - exons [ , 1] + 1)[1:i])
    }
    exons <- cbind(exons, cumuSums[1:length(cumuSums)])


    # Return a function that uses the exon matrix, strand, and start, stop positions
    return(
        function(x) {
            # Check if out-of-range
            if (x < start | x > end) { return(NULL) }

            # Determine the exon our coordinate is located in
            iEx <- which(x >= exons[ , 1] & x <= exons[ , 2])

            # Check if intron
            if(length(iEx) != 1){ return(0) }

            # Exon bases to the left (toward 5', + strand gene start) of x
            total <- exons[iEx, 3] - (exons[iEx, 2] - x)

            if (reverse) {
                #Exon bases to the right (toward 3', - strand gene start) of x
                total <- exons[nrow(exons), 3] - total + 1
            }

            return(total)
        }
    )
}
```

Sabrina-Ge:
Perhaps I should just submit another pull request

inst/scripts/.importMapCoord.R
```diff
@@ -0,0 +1,188 @@
...
+sampleFxSte(250)
+
+
+
+#===================Overall function factory prototype==========================
+
+mapCoord <- function(geneData, dropExonBoundaries = TRUE, silent = FALSE, writeLog = TRUE){
+
+    # parallelized for speed
+    library(parallel)
```
RicardoRam: Don't include library calls here. If necessary, update DESCRIPTIONS table

inst/scripts/.importMapCoord.R
```diff
@@ -0,0 +1,188 @@
...
+sampleFxSte(203)
+sampleFxSte(221)
+sampleFxSte(249)
+sampleFxSte(250)
+
+
+
+#===================Overall function factory prototype==========================
+
+mapCoord <- function(geneData, dropExonBoundaries = TRUE, silent = FALSE, writeLog = TRUE){
```
RicardoRam: ToDo: function discription

inst/scripts/.importMapCoord.R
```diff
@@ -0,0 +1,188 @@
+# importMapCoord.R
+# non-exported
+
+#==========================makeMap and Validation===============================
+
+# Helper function - generates the closure for a single gene model
+makeMap <- function(ucscMod) {
```
RicardoRam: ToDo: function description

inst/scripts/.importMapCoord.R
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    saveRDS(geneData, file = "geneData.rds")
+
+    stopCluster(cl)
+}
+
+
+
+# [END]
```
RicardoRam: ToDo: Unit Tests

hyginn:
Re. ```cumuSums```: Great that you got that to work.  But there actually **is** an R function ```cumsum()``` and that simplifies the loop to this expression:
```
exons <- cbind(exons, cumsum(exons[,2] - exons[,1] + 1))
```

Now: re. (-) strand. I think I figured it out: The trick is to convert everything to distance-from-start. Then the computation is perfectly symmetric, regardless of the strand. In fact- we don't even need to check the strand since it is implied in the start > end for all exons. I am assuming that the e.g. the (-) strand version of the "Bob-gene" would be:
```
ucscMod <- list(strand='+', start=300, end=200, exonStart=c(250,230,210), exonEnd=c(241,221,201))
```

Would you like to puzzle this out? If not, I'll just do it. (This is fun.)

Sabrina-Ge:
I will leave the problem to you! 

About the Bob-gene, though: I was looking at the UCSC data (retrieved with [this method](http://steipe.biochemistry.utoronto.ca/abc/students/index.php/User:Bagherig/Tasks/Define_best_gene-model_source#Obtaining_Data_from_UCSC)), and their ends are always greater than their starts, even when the strand is negative.

PR #86, opened 2017-04-03
SEL3D tests implemented

LindsayLiang:


inst/scripts/SEL3D.R
```diff
@@ -0,0 +1,224 @@
...
+#
+#' \code{getPdbIds} BLASTs the input cDNA sequence against PDB and returns an MT object
+#' containing the blast matches from PDB for homologous (>40% ID) sequences.
+#'
+#' Details.
+#' @section <title>: Additional explanation.
+#'
+#' @param targetSeq input sequence to be queried, must be cDNA of a length divisible by 3.
+#' @param startPos start position of the sequence, set to 1 by default, altered in recursive steps
+#' @param endPos end position of sequence, set to 0, altered to aa length in recursive steps
```
RicardoRam: startPos and endPos are parameters for your helper function getPdbIds and not for SEL3D. SEL3D and getPdbIds should both have their own function descriptions.

tests/testthat/testSel3D.R
```diff
@@ -0,0 +1,109 @@
...
+  expect_equal(result[result$pdb.id == "1DT7_X", 8], 369)
+  expect_equal(result[result$pdb.id == "1DT7_X", 9], 388)
+  expect_equal(result[result$pdb.id == "1DT7_X", 10], 1)
+  expect_equal(result[result$pdb.id == "1DT7_X", 11], 22)
+  expect_equal(result[result$pdb.id == "1DT7_X", 17], "Homo sapiens")
+  
+})
+
+
+test_that("a corrupt input does not lead to corrupted output", {
```
RicardoRam: what about tests for parameters convert and silent?

LindsayLiang: for which section?  For "a sane input gives an expected output", convert was implicitly tested with both True and False values (default was False), and for silent, how would one check the outputted messages?

RicardoRam: Under "parameter errors are correctly handled", you should have tests like

+    # if silent is not given logical input
+    expect_error(importSNV.TCGA(fMAF = testMAF, tOut, silent = 1), ".checkArgs> \"silent\" mode error: argument has mode \"numeric\" but function expects mode \"logical\".")
+    expect_error(importSNV.TCGA(fMAF = testMAF, tOut, silent = NULL), ".checkArgs> \"silent\" mode error: argument has mode \"NULL\" but function expects mode \"logical\".")
+    expect_error(importSNV.TCGA(fMAF = testMAF, tOut, silent = c(TRUE, TRUE)), ".checkArgs> \"silent\" length error: argument has length 2 but function expects length 1.")

PR #94, opened 2017-04-03
mapCoord.R + tests 

Sabrina-Ge:
`devtools::check()` runs with no errors, warnings, or notes as long as you comment out the `test_that("a sane input gives an expected output"` test from testCombine.R. (It looks like whoever updated testCombine.R or logTools.R caused the errors)

Last pull request, the function was importMapCoord.R, but I changed it to mapCoord.R to be consistent with the [task page](http://steipe.biochemistry.utoronto.ca/abc/students/index.php/BCB420_2017_Tasks/mapCoord) and since it made more sense (the function doesn't really import data from a file like the other import functions)


R/mapCoord.R
```diff
@@ -0,0 +1,170 @@
+# mapCoord.R
```
hyginn, line 1: This function won't be exported (probably).

R/mapCoord.R
```diff
@@ -0,0 +1,170 @@
...
+#' corresponding function closure that maps chromosomal coordinates to
+#' gene coordinates.
+#' The populated GeneData list is saved to an .rds file.
+#'
+#' @section Mapping function closure:
+#' The mapping function that \code{mapCoord.R} creates and populates a GeneData
+#' list which takes a chromosomal coordinate as input and returns the
+#' corresponding gene coordinate if the chromosomal coordinate occurs in an
+#' exon of the gene.
+#' If the input is in an intron, the function returns 0.
```
hyginn, line 15: What should the function do when the coordinate is in 5' or 3' UTR?

hyginn:
Code copied to the repository.

PR #110, opened 2017-04-06
GOannotation function and tests

CharlesD1ng:
Pseudocodes for GOannotation.R and testGOannotation.R!

For the sample data files, I created 3 root node and 8 other GO terms that
can build to a 3 layer tree, where:
GO:000001 is connected to biological process
GO:000002 is_obsolete=true
GO:000003 is also connected to biological process
GO:000004 is connected to molecular function
GO:000005 is connected to cellular component
GO:000006 is the child of GO:0000003
GO:000007 is the child of GO:0000006, but with no annotation
GO:000008 is the child of GO:0000007


hyginn:
Reviewed.

PR #113, opened 2017-04-06
Initial code for consense

joelostblom:
very initial...


R/consense.R
```diff
@@ -0,0 +1,53 @@
...
+    #   noLog: if FALSE, save process to an .Rdata log file#   delta: threshold for edge removal for given method
+    #   silent: if FALSE, print process to console in real time#   EGG: 'Heat Diffusion' annotated AGG, iGraph object.
+
+    #OUTPUT:
+    #   cSN: List of subnetworks resulting after edge removal and
+    #       subnetwork construction
+
+
+    #minEdgeCC = 0 #length(outputGraphs)
+    numGraphs <- length(outputGraphs)
```
hyginn, line 25: outputGraphs undefined. sN?

R/consense.R
```diff
@@ -0,0 +1,53 @@
...
+    sNEdges <- vector(mode="list", length=numGraphs)
+    sNVerts <- vector(mode="list", length=numGraphs)
+    edgePairsList <- c() #vector(mode="list", length=numGraphs)
+    for (i in 1:numGraphs){
```
hyginn, line 29: I would make a hash with keys paste()'ed from incident vertices, and just count occurrences. Our input graphs are simple, so any observation must stem from a different subnetwork. They are also directed, so you don't even have to worry about whether a-b is also b-a (Nb. this problem is usually resolved in undirected graphs by lexically sorting the incident vertices). Just dump the incident edges into a data frame for each input graph, then iterate over the rows, make keys and create a hash entry (if new) or increment its value (if it exists).

R/consense.R
joelostblom: Yes, this was supposed to be `sN`

R/consense.R
joelostblom: Thanks Boris. I was trying to create the pasted combined vertices in `edgePairs` on line 40. I was using a list, and tried to create the count table from that list, but I agree that your strategy makes more sense.

R/consense.R
joelostblom: Yes, that should have been `sN`.

